PR #7, opened 2020-01-11
Codidact skeleton

misha130:
I saw https://github.com/codidact/core/pull/6 and wanted to give this a try of my own.

This includes a folder structure, a basic MVC project that is offered by VS, a tests folder with all test projects. 

There is no ASP.Identity library, there is no EF included since it wasn't fully decided yet.

The folder structure is as follows:

- WebUI - the presentation project
- Application - All business logic of the application
- Domain - All POCOs, enums and such.
- Infrastructure - Persistance (DB), non business logic services, file management etc.


ArtOfCode-:
Forgive me, but... why do we need two skeletons?

appgurueu:
> Forgive me, but... why do we need two skeletons?

We only need a single one. But probably @misha130 was not satisfied with the current proposal?

misha130:
> We only need a single one. But probably @misha130 was not satisfied with the current proposal?

Yes. I added a few things and removed some.

Like I added a setup for unit testing/integration testing. 
I also removed things that are beyond the skeleton such as the authorization framework & EF.
Plus this basically sets the folder architecture.

ArtOfCode-:
Okay, makes sense. At this point I think it's likely we'll end up using EF Core, but since the decision isn't made yet we can keep it out for now.

ArtOfCode- (changes requested):
Looks good, other than the fact that Bootstrap is included. We've decided to avoid using Bootstrap and will use our own front-end framework instead; this'll need to be removed before we merge.

luap42:
I saw, that you are aware of the repository of our design framework Co-Design (design.codidact.org) already. I am not an expert in C#, but would it be possible to somehow download and include the following file (which is always linking to the latest version of CD) during build:

https://raw.githubusercontent.com/codidact/co-design/master/dist/codidact.css

Even better would be integration with SCSS, which would allow sites to override styles, but this shouldn't be/doesn't have to be/isn't part of a skeleton IMO.

ArtOfCode- (dismissed):

misha130:
> I saw, that you are aware of the repository of our design framework Co-Design (design.codidact.org) already. I am not an expert in C#, but would it be possible to somehow download and include the following file (which is always linking to the latest version of CD) during build:
> 
> https://raw.githubusercontent.com/codidact/co-design/master/dist/codidact.css
> 
> Even better would be integration with SCSS, which would allow sites to override styles, but this shouldn't be/doesn't have to be/isn't part of a skeleton IMO.

Done

A few points:
Had to download the css/css.map to the project because the link you provided just gave me text/plain rather text/css so Chrome blocked it.
Had to also download the fonts into the project. 
I also took the co-design docs and made a simple page with a menu.
Its actually ready to be designed from this point IMO.

src/WebUI/Views/Shared/_Layout.cshtml
```diff
@@ -1,48 +1,52 @@
...
-<html lang="en">
+<html lang="en" dir="ltr">
 <head>
-    <meta charset="utf-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>@ViewData["Title"]</title>
-    <link rel="stylesheet" href="~/lib/bootstrap/dist/css/bootstrap.min.css" />
+    <script src="~/lib/jquery/dist/jquery.min.js"></script>
```
luap42: Note, that scripts should be loaded asynchronous or deferred according to the coding standards :
https://github.com/codidact/core/wiki/Code-standards:-HTML#25-links-and-scripts

misha130: Removed jQuery.
Set site script as async.

ArtOfCode- (approved):

ArtOfCode-:
Merging. We need to get developing.

PR #13, opened 2020-01-15
CSS cleanup

luap42:
I'm working with ASP.NET Core for the first time, here, so please give helpful critiques.

This PR updates the version of Co-Design, removes obsolete content of site.css, fixes some issues with the header (missing or wrong link targets), adds a header slide (mobile nav) JavaScript and restyles the footer witb CD.


src/WebUI/Views/Shared/_Layout.cshtml
```diff
@@ -30,17 +30,17 @@
...
         <div class="header-slide--separator"></div>
-        <a class="header-slide--item" href="https://github.com/codidact/co-design">on GitHub</a>
+        <a class="header-slide--item" href="https://github.com/codidact/core">on GitHub</a>
     </aside>
     <main class="container has-padding-4">
         @RenderBody()
     </main>
 
-    <footer class="border-top footer text-muted">
+    <footer class="has-padding-4 has-border-top-style-solid has-border-top-width-1 has-border-color-tertiary-050">
```
ArtOfCode-, line 43: Not sure I like this. Can we take the effect of all these classes on `footer` and write them into a default style for `footer`?

src/WebUI/wwwroot/js/site.js
```diff
@@ -1,4 +1,25 @@
...
+        // Position header slide appropriately relative to
+        // trigger.
+        const rect = this.getBoundingClientRect();
+        hs.style.top = (rect.top + rect.height) + "px";
+        hs.style.right = (document.body.clientWidth - rect.right) + "px";
+
+        // Prevent navigation
+        e.preventDefault();
+    });
+}
```
ArtOfCode-, line 25: It's struck me that this means navigation doesn't work properly without JS enabled, which goes against the whole progressive-enhancement shtick: core functionality, including navigation, should work fully and correctly without JS enabled. It can be _enhanced_ with JS, but should work properly without. I'm not sure whether this is something that can/should be fixed here, or in Co-Design - probably the latter, I guess?

src/WebUI/Views/Shared/_Layout.cshtml
luap42: That's definitely possible and probably a good idea. However we'd need to review, how our footer should look like (minimal like here, large as on SE...) before doing so.

src/WebUI/wwwroot/js/site.js
luap42: It depends. Only mobile nav needs JavaScript. I think it is less common not to have JS on mobile, but you're right, that it should still work.

We could built some complex CSS-only solution using the URL hashes and the :target pseudoselector. Or we add a noscript element, which always shows the nav.

ArtOfCode- (approved):

PR #20, opened 2020-01-18
Add runtime migration execution

### LINKED ISSUE #17: Missing Migration execution at Runtime
As per https://github.com/codidact/core/pull/15/commits/c7dd9d8c53cc0c8fc724a9ac70e1ad284eb94c33 currently the execution of the migration is execute via a cli command.

Rather than that, this line of code should be added somewhere on start:
`dbContext.Database.Migrate()`

Exception handling for this should also be implemented since there is no proper connection string.

benjamin-allen:
Closes #17.

Migrations are applied as part of the `Startup` `Configure` method.

### DIFF
--- a/src/WebUI/Startup.cs
+++ b/src/WebUI/Startup.cs
@@ -7,6 +7,7 @@ using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
 
 namespace Codidact.WebUI
 {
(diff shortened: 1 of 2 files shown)


src/WebUI/Startup.cs
```diff
@@ -53,6 +57,29 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
                     pattern: "{controller=Home}/{action=Index}/{id?}");
             });
+
+            ApplyDatabaseMigrations(app, logger);
+        }
+
+        // Applies database migrations; won't cause any changes if the database is up-to-date.
+        private void ApplyDatabaseMigrations(IApplicationBuilder app, ILogger logger)
+        {
+            using (var serviceScope = app.ApplicationServices.GetRequiredService<IServiceScopeFactory>().CreateScope())
```
asynts, line 67:
```suggestion
            using (var serviceScope = app.ApplicationServices.CreateScope())
```
This seems to be enough, at least on my machine.

ArtOfCode- (approved):

PR #23, opened 2020-01-19
WIP: Add middleware that normalizes subdomain vs. path url scheme.

asynts:
This is a first step for #22. I added some middleware that rewrites as follows:

~~~none
https://codidact.com         -> https://codidact.com
https://foo.codidact.com     -> https://codidact.com/community/foo/
https://foo.codidact.com/bar -> https://codidact.com/community/foo/bar
~~~

*The following has been edited.*

The following things still need to be addressed: (not in this pull request though)

 1. `foo.codidact.com` and `codidact.com/community/foo` are both accepted, maybe redirect
     to the subdomain version when enabled?

PR #27, opened 2020-01-22
Salvage the configuration classes from #23.

asynts:
It took me a few hours to figure out how all this configuration stuff is supposed to be used. Codidact will need configuration options at some point, it would be a shame if someone had to invest that time again in a week or so.


ArtOfCode- (dismissed):

asynts:
> It took me a few hours to figure out how all this configuration stuff is supposed to be used. 

I think that was only because I wasn't familiar with the framework. That doesn't mean that it's useless, however, this would collide with #29 and cause more trouble than necessary.

PR #29, opened 2020-01-25
IdentityServer & Identity.net implementation

### LINKED ISSUE #21: Implement IdentityServer & Identity.NET
As per https://forum.codidact.org/t/what-kind-of-authorization-framework-library-should-we-use/642/7 it has been decided upon IdentityServer/Identity.NET

The following adjustments should be:
* Add IdentityServer and Identity configuration to Startup: 
```
   app.UseAuthentication();
   app.UseIdentityServer();
   app.UseAuthorization();
```
* Change ApplicationDbContext to inherit IdentityDbContext
* Create an ApplicationUser that inherits IdentityUser
* Add IdentityServer to DI in the infrastructure module:

```
   services.AddIdentityServer()
        .AddApiAuthorization<ApplicationUser, ApplicationDbContext>();

```

* Set member id on create/delete/update
* Integration tests for both WebUI/Infrastructure

misha130:
Closes #21 

UI for IdentityServer taken from https://github.com/IdentityServer/IdentityServer4/tree/master/samples/Quickstarts/6_AspNetIdentity/src/IdentityServerAspNetIdentity

Currently no tests on Auth.

Includes a claim for memberId which every entity gets on insert/update/delete


src/Application/Common/Interfaces/ICurrentUserService.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        /// <summary>
+        /// Retreives the member id of the current user.
+        /// </summary>
+        public long GetMemberId();
+
+        /// <summary>
+        /// Retreives the user id of the current user.
+        /// </summary>
+        public string GetUserId();
```
ArtOfCode-, line 13: Just so I'm clear: what's the difference between `MemberId` and `UserId`?

src/Auth/Quickstart/Account/AccountController.cs
```diff
@@ -0,0 +1,346 @@
+// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+
```
ArtOfCode-: Spurious extra blank line here

src/Auth/Quickstart/Account/AccountController.cs
```diff
@@ -0,0 +1,346 @@
...
+        /// Handle logout page postback
+        /// </summary>
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<IActionResult> Logout(LogoutInputModel model)
+        {
+            // build a model so the logged out page knows what to display
+            var vm = await BuildLoggedOutViewModelAsync(model.LogoutId);
+
+            if (User?.Identity.IsAuthenticated == true)
```
ArtOfCode-: All navigations after a safe navigation operator must also be safetied: `User?.Identity?.IsAuthenticated`.

If `User?.Identity` returns `null`, then you get an exception on the call to `IsAuthenticated` (because `User?.Identity.IsAuthenticated == (null).IsAuthenticated`, so that call must also use the SNO.

src/Auth/Quickstart/Account/AccountController.cs
```diff
@@ -0,0 +1,346 @@
...
+                .Where(x => x.DisplayName != null ||
+                            (x.Name.Equals(AccountOptions.WindowsAuthenticationSchemeName, StringComparison.OrdinalIgnoreCase))
+                )
+                .Select(x => new ExternalProvider
+                {
+                    DisplayName = x.DisplayName,
+                    AuthenticationScheme = x.Name
+                }).ToList();
+
+            var allowLocal = true;
```
ArtOfCode-: Magic boolean! Should this instead be a configuration option?

src/Auth/Quickstart/Account/AccountController.cs
```diff
@@ -0,0 +1,346 @@
...
+            var vm = new LoggedOutViewModel
+            {
+                AutomaticRedirectAfterSignOut = AccountOptions.AutomaticRedirectAfterSignOut,
+                PostLogoutRedirectUri = logout?.PostLogoutRedirectUri,
+                ClientName = string.IsNullOrEmpty(logout?.ClientName) ? logout?.ClientId : logout?.ClientName,
+                SignOutIframeUrl = logout?.SignOutIFrameUrl,
+                LogoutId = logoutId
+            };
+
+            if (User?.Identity.IsAuthenticated == true)
```
ArtOfCode-: Another chained SNO needed here: `User?.Identity?.IsAuthenticated`

src/Auth/Quickstart/Account/AccountOptions.cs
```diff
@@ -0,0 +1,25 @@
...
+
+        public static bool ShowLogoutPrompt = true;
+        public static bool AutomaticRedirectAfterSignOut = false;
+
+        // specify the Windows authentication scheme being used
+        public static readonly string WindowsAuthenticationSchemeName = Microsoft.AspNetCore.Server.IISIntegration.IISDefaults.AuthenticationScheme;
+        // if user uses windows auth, should we load the groups from windows
+        public static bool IncludeWindowsGroups = false;
+
+        public static string InvalidCredentialsErrorMessage = "Invalid username or password";
```
ArtOfCode-: This may not be one to do in this PR, but we'll need to move this to an i18n strings file at some point so we're not using magic English constants.

src/Auth/Quickstart/Account/ExternalController.cs
```diff
@@ -0,0 +1,313 @@
...
+            if (user == null)
+            {
+                // this might be where you might initiate a custom workflow for user registration
+                // in this sample we don't show how that would be done, as our sample implementation
+                // simply auto-provisions new external user
+                user = await AutoProvisionUserAsync(provider, providerUserId, claims);
+            }
+
+            // this allows us to collect any additonal claims or properties
+            // for the specific prtotocols used and store them in the local auth cookie.
```
ArtOfCode-: Typo in `protocols`

src/Auth/Quickstart/Account/ExternalController.cs
```diff
@@ -0,0 +1,313 @@
...
+            var name = claims.FirstOrDefault(x => x.Type == JwtClaimTypes.Name)?.Value ??
+                claims.FirstOrDefault(x => x.Type == ClaimTypes.Name)?.Value;
+            if (name != null)
+            {
+                filtered.Add(new Claim(JwtClaimTypes.Name, name));
+            }
+            else
+            {
+                var first = claims.FirstOrDefault(x => x.Type == JwtClaimTypes.GivenName)?.Value ??
+                    claims.FirstOrDefault(x => x.Type == ClaimTypes.GivenName)?.Value;
```
ArtOfCode-: Do we need this? We don't need to collect "real" first and last name - if this is just handling external IdP's, that's fine, but we don't need to store the data on our native user object.

src/Auth/Quickstart/Consent/ConsentOptions.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+
+namespace Codidact.Auth.Quickstart.UI
+{
+    public class ConsentOptions
+    {
+        public static bool EnableOfflineAccess = true;
```
ArtOfCode-: More strings in this file that will need to be moved to i18n at some point.

src/Auth/SeedDatabase.cs
```diff
@@ -0,0 +1,34 @@
...
+                }
+            };
+            using var serviceScope = app.ApplicationServices.GetRequiredService<IServiceScopeFactory>().CreateScope();
+            using var userManager = serviceScope.ServiceProvider.GetService<UserManager<ApplicationUser>>();
+            foreach (var entity in toCreateApplicationUsers)
+            {
+                var exists = await userManager.FindByNameAsync(entity.UserName);
+                if (exists == null)
+                {
+                    await userManager.CreateAsync(entity, "Aa123456!");
```
ArtOfCode-: A better way of doing this might be to randomly generate the password and print it to the console.

src/Auth/Startup.cs
```diff
@@ -0,0 +1,82 @@
...
+                options.Events.RaiseSuccessEvents = true;
+            })
+                .AddDeveloperSigningCredential()
+                .AddInMemoryClients(Configuration.GetSection("IdentityServer:Clients"))
+                .AddInMemoryIdentityResources(Config.GetIdentityResources())
+                .AddInMemoryApiResources(Configuration.GetSection("IdentityServer:ApiResources"))
+                // this adds the operational data from DB (codes, tokens, consents)
+                .AddAspNetIdentity<ApplicationUser>()
+                .AddProfileService<CustomProfileService>()
+                ;
```
ArtOfCode-: Semicolon should be on the previous line with the end of the statement.

src/Auth/Views/Account/Logout.cshtml
```diff
@@ -0,0 +1,21 @@
+﻿@model LogoutViewModel
+
+<div class="logout-page">
+    <div class="page-header">
+        <h1>Logout</h1>
+    </div>
+
+    <div class="row">
+        <div class="col-sm-6">
+            <p>Would you like to logout of IdentityServer?</p>
```
ArtOfCode-: Should probably be "Would you like to log out of (ClientName)?" rather than IdentityServer. Users will recognise the former, not the latter.

src/Auth/Views/Consent/Index.cshtml
```diff
@@ -0,0 +1,82 @@
...
+    </div>
+
+    <div class="row">
+        <div class="col-sm-8">
+            <partial name="_ValidationSummary" />
+            
+            <form asp-action="Index" class="consent-form">
+                <input type="hidden" asp-for="ReturnUrl" />
+
+                <div>Uncheck the permissions you do not wish to grant.</div>
```
ArtOfCode-: At least the "Personal Information" scope _must_ be default-unchecked, otherwise we hit GDPR.

src/Auth/Views/Device/UserCodeConfirmation.cshtml
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    <div class="row">
+        <div class="col-sm-8">
+            <partial name="_ValidationSummary" />
+
+            <form asp-action="Callback" class="consent-form">
+                <input asp-for="UserCode" type="hidden" value="@Model.UserCode" />
+
+                <div>Uncheck the permissions you do not wish to grant.</div>
```
ArtOfCode-: Likewise, personal information should be default-unchecked.

src/Auth/Views/Grants/Index.cshtml
```diff
@@ -0,0 +1,79 @@
...
+                            <span class="expires">Expires:</span> @grant.Expires.Value.ToString("yyyy-MM-dd")
+                        </div>
+                    }
+                    @if (grant.IdentityGrantNames.Any())
+                    {
+                        <div>
+                            <div class="granttype">Identity Grants</div>
+                            <ul>
+                                @foreach (var name in grant.IdentityGrantNames)
+                        {
```
ArtOfCode-: Indentation error

src/Auth/Views/Home/Index.cshtml
```diff
@@ -0,0 +1,39 @@
+@{
```
ArtOfCode-: Is this page necessary? Seems to be more developer-oriented than user-facing, what with the source and example links.

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
+﻿body {
```
ArtOfCode-, line 1: Whole file needs review and updating to the CSS standards. Specifically, newlines between rules are missing. If this is a compiled file generated from the .less, it should be minified.

src/Auth/wwwroot/css/site.less
```diff
@@ -0,0 +1,116 @@
+﻿body {
```
ArtOfCode-, line 1: Are we stuck with LESS here, or could we use SCSS instead?

src/Auth/wwwroot/js/signout-redirect.js
```diff
@@ -0,0 +1,6 @@
+﻿window.addEventListener("load", function () {
```
ArtOfCode-: Prefer `() => {` syntax (ES6) over `function () {`.

src/Auth/wwwroot/js/signout-redirect.js
```diff
@@ -0,0 +1,6 @@
...
+    var a = document.querySelector("a.PostLogoutRedirectUri");
```
ArtOfCode-: Use `const` rather than `var`.

src/Auth/wwwroot/js/site.js
```diff
@@ -0,0 +1,25 @@
+﻿/*
+ * Header slides
+ ----------------
+ * This code powers the "header slides", which are at the core of mobile nav
+*/
+
+const headerSlideTriggers = document.querySelectorAll("[data-trigger-header-slide]");
+
+for (let i = 0; i < headerSlideTriggers.length; i++) {
+    headerSlideTriggers[i].addEventListener("click", function (e) {
```
ArtOfCode-, line 10: Prefer `(e) => {`

src/Infrastructure/Identity/CurrentUserService.cs
```diff
@@ -0,0 +1,42 @@
...
+            _httpContextAccessor = httpContextAccessor;
+        }
+
+        /// <summary>
+        /// Retreives the user id of the current user.
+        /// </summary>
+        public string GetUserId()
+        {
+            return _httpContextAccessor.HttpContext?.User?.FindFirstValue(ClaimTypes.NameIdentifier);
+
```
ArtOfCode-: Extra blank line

src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -238,6 +238,307 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                        .HasColumnType("integer");
+
+                    b.Property<string>("ConcurrencyStamp")
+                        .IsConcurrencyToken()
+                        .HasColumnName("concurrency_stamp")
+                        .HasColumnType("text");
+
+                    b.Property<string>("Email")
+                        .HasColumnName("email")
+                        .HasColumnType("character varying(256)")
```
ArtOfCode-: Emails have a max length of 320 characters - 256 isn't enough for an email field.

src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -238,6 +238,307 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                        .HasColumnName("email_confirmed")
+                        .HasColumnType("boolean");
+
+                    b.Property<bool>("LockoutEnabled")
+                        .HasColumnName("lockout_enabled")
+                        .HasColumnType("boolean");
+
+                    b.Property<DateTimeOffset?>("LockoutEnd")
+                        .HasColumnName("lockout_end")
+                        .HasColumnType("timestamp with time zone");
```
ArtOfCode-, line 261: Does this need `with time zone`? We should be storing datetimes as UTC in the database, regardless of application or client time zone.

src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -238,6 +238,307 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+
+                    b.Property<long?>("MemberId")
+                        .HasColumnName("member_id")
+                        .HasColumnType("bigint");
+
+                    b.Property<string>("NormalizedEmail")
+                        .HasColumnName("normalized_email")
+                        .HasColumnType("character varying(256)")
```
ArtOfCode-, line 269: Likewise, 320 minimum for an email field.

src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs
```diff
@@ -238,6 +238,307 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
+                    b.Property<string>("NormalizedUserName")
+                        .HasColumnName("normalized_user_name")
+                        .HasColumnType("character varying(256)")
+                        .HasMaxLength(256);
+
+                    b.Property<string>("PasswordHash")
+                        .HasColumnName("password_hash")
+                        .HasColumnType("text");
+
+                    b.Property<string>("PhoneNumber")
```
ArtOfCode-: We don't want to be collecting phone numbers. Is this required for IdentityServer, or can we remove this (and its Confirmed sibling, next)?

src/WebUI/Startup.cs
```diff
@@ -25,6 +26,32 @@ public void ConfigureServices(IServiceCollection services)
...
+                options.DefaultSignInScheme = "cookie";
+                options.DefaultChallengeScheme = "oidc";
+            })
+            .AddCookie("cookie")
+            .AddOpenIdConnect("oidc", options =>
+            {
+                options.Authority = "http://localhost:5000";
+                options.RequireHttpsMetadata = false; // dev only
+                options.ClientId = "codidact_client";
+                options.ClientSecret = "acf2ec6fb01a4b698ba240c2b10a0243";
```
ArtOfCode-: Is this okay to be kept in public Git history, or should it be made a configurable value?

src/WebUI/Views/Shared/_Layout.cshtml
```diff
@@ -34,6 +42,14 @@
         <a class="header-slide--item" href="#">Placeholder 2</a>
         <a class="header-slide--item" href="#">Placeholder 3</a>
         <div class="header-slide--separator"></div>
+        @if (User.Identity.IsAuthenticated)
```
ArtOfCode-, line 45: Should this be the other way round? If the user is authenticated, we should show a sign-out link; if not, it should be a sign-in.

tests/Infrastructure.IntegrationTests/Persistence/ApplicationDbContextTests.cs
```diff
@@ -2,19 +2,20 @@
...
+namespace Codidact.Infrastructure.IntegrationTests.Persistence
 {
     public class ApplicationDbContextTests
     {
         private readonly ApplicationDbContext _sutContext
             = new ApplicationDbContext(
                 new DbContextOptionsBuilder<ApplicationDbContext>()
                 .UseInMemoryDatabase(Guid.NewGuid().ToString())
-                .Options
+                .Options, new CurrentUserServiceMock()
```
ArtOfCode-: This syntax is unclear. I'd suggest putting `new DbContextOptionsBuilder<ApplicationDBContext>().UseInMemoryDatabase(Guid.NewGuid().ToString()).Options` on one line, then `new CurrentUserServiceMock()` on the next with the closing parenthesis.

ArtOfCode- (changes requested):

src/Application/Common/Interfaces/ICurrentUserService.cs
misha130: So you have this table from the identity.net which is created automatically from as part of the library among like 7 other tables like AspNetRoles, AspNetUserClaims, etc:

```
CREATE TABLE public."AspNetUsers" (
	id text NOT NULL,
	user_name varchar(256) NULL,
	normalized_user_name varchar(256) NULL,
	email varchar(256) NULL,
	normalized_email varchar(256) NULL,
	email_confirmed bool NOT NULL,
	password_hash text NULL,
	security_stamp text NULL,
	concurrency_stamp text NULL,
	phone_number text NULL,
	phone_number_confirmed bool NOT NULL,
	two_factor_enabled bool NOT NULL,
	lockout_end timestamptz NULL,
	lockout_enabled bool NOT NULL,
	access_failed_count int4 NOT NULL,
	member_id serial NULL
        CONSTRAINT pk_users PRIMARY KEY (id)
);
```
And the members table:
```
CREATE TABLE public.members (
	id bigserial NOT NULL,
	display_name varchar(100) NOT NULL,
	bio text NULL,
	email varchar(320) NOT NULL,
	"location" text NULL,
	is_from_stack_exchange bool NOT NULL DEFAULT false,
	stack_exchange_id int8 NULL,
	is_email_verified bool NOT NULL,
	is_suspended bool NOT NULL DEFAULT false,
	create_date_at timestamp NOT NULL DEFAULT '0001-01-01 00:00:00'::timestamp without time zone,
	created_by_member_id int8 NOT NULL DEFAULT 0,
	deleted_by_member_id int8 NULL,
	is_deleted bool NOT NULL DEFAULT false,
	last_modified_by_member_id int8 NULL,
	deleted_at timestamp NULL,
	last_modified_at timestamp NULL,
	stack_exchange_last_imported_at timestamp NULL,
	stack_exchange_validated_at timestamp NULL,
	suspension_end_at timestamp NULL,
	CONSTRAINT pk_members PRIMARY KEY (id)
);
```
I made them connected so the User can be connected to the member. Now generally the Authentication only handles the ApplicationUser and not the Member. The member is part of the application and not the Authorization.

This is not me making things up, this is just more or less using default configuration. 
What I could do though is remove email & email_verified from member.

src/WebUI/Startup.cs
misha130: I'll move this to the app.settings file but leave it there as an example. 
Just don't use this specific secret in production

src/Auth/Quickstart/Account/AccountController.cs
misha130: The first one that breaks the chain is the value returned. So the whole expression becomes the null. So User?.Identity.IsAuthenticated if user is null just returns null and never checks the other ones.

(fiddler.net sucks because it doesn't take .? at all)

src/Auth/wwwroot/css/site.less
misha130: I'll look into it and how it even compiles. I didn't check at all

src/Auth/Quickstart/Account/AccountController.cs
ArtOfCode-: Huh, TIL. Never knew that about C#'s SNO - thanks!

appgurueu (commented):
Is it really necessary to prefix the tables with `AspNet` ?

src/Auth/Quickstart/Account/AccountOptions.cs
misha130: True

src/Auth/Common/AccountOptions.cs
```diff
@@ -1,25 +1,18 @@
 ﻿// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
```
appgurueu, line 1: This is irritating me. Are you sure we won't have any licensing issues?

src/Auth/Controllers/AccountController.cs
```diff
@@ -1,10 +1,11 @@
 // Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
```
appgurueu, line 1: Same as above.

src/Auth/Models/ErrorViewModel.cs
```diff
@@ -1,10 +1,9 @@
 ﻿// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
```
appgurueu, line 1: ...

appgurueu (changes requested):

misha130:
> Is it really necessary to prefix the tables with `AspNet` ?

This configured but not directly. I'll have to add:
 ```
        modelBuilder.Entity<ApplicationUser>().ToTable("User");
        modelBuilder.Entity<IdentityRole>().ToTable("Role");
        modelBuilder.Entity<IdentityUserRole>().ToTable("UserRole");
        modelBuilder.Entity<IdentityUserClaim>().ToTable("UserClaim");
        modelBuilder.Entity<IdentityUserLogin>().ToTable("UserLogin");
```
Yay/Nay?

File: src/Auth/Models/ErrorViewModel.cs (hunk 32 above)
Comment (misha130) [reply src/Auth/Models/ErrorViewModel.cs:1]: I asked in discord @ArtOfCode- and he said its ok. I personally am not sure what to do about these

**Comment by appgurueu** [comment]
Definitely yay. See [this question on SE](https://softwareengineering.stackexchange.com/questions/191929/the-problems-with-avoiding-smurf-naming-classes-with-namespaces). And most importantly, the "AspNet" prefix doesn't convey any meaning; it's just unnecessary payload.

File: src/Auth/Models/ErrorViewModel.cs (hunk 32 above)
Comment (appgurueu) [reply src/Auth/Models/ErrorViewModel.cs:1]: The [Apache 2](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0)) license is not what is bothering me much, but rather the contradicting "all rights reserved" statement.

**Comment by misha130** [comment]
> Definitely yay.

Resolved.

Edit:
Had to rename to the naming convention: users, roles, user_roles, user_claims, user_logins

File: src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs (hunk 26 above)
Comment (misha130) [reply src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs]: Removed

File: src/Auth/SeedDatabase.cs (hunk 10 above)
Comment (misha130) [reply src/Auth/SeedDatabase.cs]: That is a good idea

File: src/Application/Common/Interfaces/ICurrentUserService.cs (hunk 1 above)
Comment (misha130) [reply src/Application/Common/Interfaces/ICurrentUserService.cs:13]: Removed Email in https://github.com/codidact/core/pull/29/commits/95cfc8bec9fb593a1eed551c6a89723beef041b7

File: tests/Infrastructure.IntegrationTests/Persistence/ApplicationDbContextTests.cs (hunk 29 above)
Comment (misha130) [reply tests/Infrastructure.IntegrationTests/Persistence/ApplicationDbContextTests.cs]: Fixed in https://github.com/codidact/core/pull/29/commits/7eb646f8cbcb25a64a7d085f69f710389e1b3f9c

File: src/Auth/Startup.cs (hunk 11 above)
Comment (misha130) [reply src/Auth/Startup.cs]: Resolved in https://github.com/codidact/core/pull/29/commits/7eb646f8cbcb25a64a7d085f69f710389e1b3f9c

File: src/Auth/Views/Account/Logout.cshtml (hunk 12 above)
Comment (misha130) [reply src/Auth/Views/Account/Logout.cshtml]: I think this is impossible because the token does not hold the client name with which it has been authenticated. I am 90% sure. It might be possible on login but not log out. I might be wrong

File: src/Auth/wwwroot/css/site.less (hunk 18 above)
Comment (misha130) [reply src/Auth/wwwroot/css/site.less:1]: This file is just the source. I think they just compiled the LESS manually and then added a .css.min and a .css min to the project so yes you can switch to SCSS but there is just no compiler whatsoever.

File: src/Infrastructure/Identity/CurrentUserService.cs (hunk 22 above)
Comment (misha130) [reply src/Infrastructure/Identity/CurrentUserService.cs]: Fixed in https://github.com/codidact/core/pull/29/commits/55abf50aba5742c2e43c128da7af67e2c8b2d41e

File: src/Auth/wwwroot/js/signout-redirect.js (hunk 19 above)
Comment (misha130) [reply src/Auth/wwwroot/js/signout-redirect.js]: There is no Babel here to convert the ES6 syntax in to ES5/ES3 supported syntax. There won't be old browser support.

File: src/Auth/Models/ErrorViewModel.cs (hunk 32 above)
Comment (ranolfi) [reply src/Auth/Models/ErrorViewModel.cs:1]: Copyright belongs to whoever wrote the code, regardless of whether this notice is present in the source code or not.*

It also applies only to the code originally written by these authors.

So, keeping the copyright notice is not only "ok", but also the sane thing to do.

(*) Except in cases where the author has explicitly [waived all copyrights](https://en.wikipedia.org/wiki/Anti-copyright_notice) by releasing the software into the [public domain](https://en.wikipedia.org/wiki/Public-domain_software), in jurisdictions that allow that.

Comment (misha130) [reply src/Auth/Models/ErrorViewModel.cs:1]: How much would the code have to change for the licensing to be removed?

Comment (ranolfi) [reply src/Auth/Models/ErrorViewModel.cs:1]: As long as there is one line of code written by that person, the copyright notice is due. And I do believe that even the file name by itself also counts. Perhaps @ArtOfCode- can confirm.

In any case: What we're seeing here is a *copyright notice*, not a license statement.

When releasing software, the copyright notice serves to identify the author who made it available to the public. This particular copyright notice belongs to Identity Server 4, which is released under the Apache 2.0 license. We, and future contributors, *should* know who licensed these files under the terms of that license, for various reasons. The notice should remain there. Legally, it would make no difference; however, it's an actual requirement of the Apache 2.0 license to have the notice in every file distributed under it. (see: https://opensource.stackexchange.com/q/6909/3988)

"All rights reserved" means, mainly, two things: (1) the author(s) is free to release newer versions of the software under a different license of their choice; and (2) they have the right to sue anyone who infringes their license terms (which is expected indeed).

Furthermore, I believe contributors can add their own copyright notices in addition to the one(s) already there. Can't state that for sure, though. More info needed.

**Comment by asynts** [comment]
I don't understand why we need the Identity Server stuff at all. If I understand it correctly, the idea is that you have multiple websites and a single centralized server for authentication (the identity server).

Do we plan on hosting multiple sites? For me it just seems like code bloat and unnecessary complexity.

**Comment by misha130** [comment]
> I don't understand why we need the Identity Server stuff at all. If I understand it correctly, the idea is that you have multiple websites and a single centralized server for authentication (the identity server).

You can implement external auths (google, etc) easier.
You can have multiple type of authentications on one server.
You have a better separation of responsibility. 
We will have an API (for mobile and developers) that is somewhat separate from the MVC project.
If we do decide to separate the instances to somewhat different servers this won't be a problem since there is still one authentication server. 

Its actually not that big of a bloat too. There is one controller and like 3 html files. 
Most of the code is basically in a black box though so yea that is a downside.

**Comment by asynts** [comment]
> You can implement external auths (google, etc) easier.

I haven't read the whole thing, [but this seems to be possible without an identity server][1].

  > You can have multiple type of authentications on one server.

Again, [this seems to be possible without an identity server][2].

  > We will have an API (for mobile and developers) that is somewhat separate from the MVC project.

I don't think that is, or should be, in the minimum viable product.

  > Its actually not that big of a bloat too.

There is yet another project. We already have "Application", "Infrastructure", "Domain" and "WebUI". At some point it becomes difficult to keep track of where the code is.

Personally, this just fells "messy".

  [1]: https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/google-logins?view=aspnetcore-3.1
  [2]: https://docs.microsoft.com/en-us/aspnet/core/security/authentication/social/?view=aspnetcore-3.1&tabs=visual-studio-code#multiple-authentication-providers

**Comment by misha130** [comment]
> Again, this seems to be possible without an identity server.

You are right, both of these things are possible. The question is how much of it is configurable and how.

> I don't think that is, or should be, in the minimum viable product.

The structure should be ready for this so you won't have to rewrite half the project to include it or have duplicate code all over the place.

> There is yet another project. We already have "Application", "Infrastructure", "Domain" and "WebUI". At some point it becomes difficult to keep track of where the code is.
> 
> Personally, this just fells "messy".

I really normally have these 4 in any project just because of how orderly it makes things. When the project grows too much you have too many files in one folder it all becomes a huge mess. So these 4 projects are a precaution for the future. Furthermore, I promise you these will be the only 3 library class projects that will be in the solution.

**Comment by asynts** [comment]
I think I missed your main, point:

  > If we do decide to separate the instances to somewhat different servers this won't be a problem since there is still one authentication server.

Still, I don't see why this can't be done later. The difference seems to be mostly `AddOpenIdConnect` and a few tweaks here and there.

**Comment by misha130** [comment]
> I think I missed your main, point:
> 
> > If we do decide to separate the instances to somewhat different servers this won't be a problem since there is still one authentication server.
> 
> Still, I don't see why this can't be done later. The difference seems to be mostly `AddOpenIdConnect` and a few tweaks here and there.

You are right on your points but you are missing the forest for the trees. 
Your main argument is that the Presentation (WebUI) project should hold more responsibility and take care of the authorization. Well if you ask me should something have more responsibility or less I would also go with the less.

I can't even imagine a situation where down the line we might say "oh we should probably have the auth in the webui to make it more compact". I can certainly imagine a situation where we might say "oh we should separate the authorization from the webui"

File: src/Auth/SeedDatabase.cs (hunk 10 above)
Comment (misha130) [reply src/Auth/SeedDatabase.cs]: Fixed in https://github.com/codidact/core/pull/29/commits/855206b88160c5281b1922dd75e696943d432a35
But its just a simple GUID. The person running an instance should be reasonable and change his password.
I don't think its a good idea to hold the code that generates passwords in the code base because then people will think its safe and won't be encouraged to change their passwords.

**Comment by asynts** [comment]
> I can't even imagine a situation where down the line we might say "oh we should probably have the auth in the webui to make it more compact". I can certainly imagine a situation where we might say "oh we should separate the authorization from the webui"

I guess that is an experience thing, I work on smaller projects normally where it is possible to change stuff like this without trouble.

File: src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs (hunk 24 above)
Comment (misha130) [reply src/Infrastructure/Persistence/Migrations/ApplicationDbContextModelSnapshot.cs:261]: I checked https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.identity.identityuser-1.lockoutend?view=aspnetcore-3.1
It says that it saves it as UTC.

I am just not sure if this will create a bug or not. EF automatically maps DateTimeOffset to timestamp with time zone. I can configure it to no time zone but I fear it will have unforeseen consequences.
I tried building a test for this but then realized I can't do in memory testing since I need the behavior of the schema tested.

File: src/WebUI/Startup.cs (hunk 27 above)
Comment (misha130) [reply src/WebUI/Startup.cs]: Fixed in https://github.com/codidact/core/pull/29/commits/91a0e356feb425644914546816d9229b1356ad61

File: src/Auth/Views/Account/Logout.cshtml (hunk 12 above)
Comment (ArtOfCode-) [reply src/Auth/Views/Account/Logout.cshtml]: In that case... well, what does this button actually _do_? Does it log you out only of our platform, or out of external IdPs as well?

File: src/Auth/wwwroot/css/site.less (hunk 18 above)
Comment (ArtOfCode-) [reply src/Auth/wwwroot/css/site.less:1]: Gotcha. In that case, let's keep the CSS but update it to our standards. The .less and the .min.css can go.

File: src/Auth/wwwroot/js/signout-redirect.js (hunk 19 above)
Comment (ArtOfCode-) [reply src/Auth/wwwroot/js/signout-redirect.js]: ...mutter mutter old browsers

File: src/Auth/Models/ErrorViewModel.cs (hunk 32 above)
Comment (ArtOfCode-) [reply src/Auth/Models/ErrorViewModel.cs:1]: "All rights reserved" is not contradictory to a license grant. It's not a _necessary_ statement, but it's also not wrong. Keeping these license headers in is the generally accepted way to go.

File: src/Auth/Views/Account/Logout.cshtml (hunk 12 above)
Comment (misha130) [reply src/Auth/Views/Account/Logout.cshtml]: Just "our platform"
```

      // delete local authentication cookie
       await _signInManager.SignOutAsync();
```

File: src/Auth/wwwroot/css/site.css [hunk 33]
```diff
@@ -0,0 +1,82 @@
...
+  margin-top: 65px;
+}
+.navbar-header {
```
mattjbrent, line 5: We use backwards BEM (not sure why). If this is two components then it should be navbar--header. Where the navbar component contains a header subcomponent

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+  position: relative;
+  top: -4px;
+}
+.navbar-brand > .icon-banner {
```
mattjbrent, line 10: same as above RE nested components using `--`.

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
+﻿body {
+  margin-top: 65px;
+}
```
mattjbrent, line 3: Add empty line between all rulesets. Also ensure tab is 4 whitespace and not tab (havent pulled the file)

src/Auth/wwwroot/css/site.css
ranolfi: Should eventually be ported to SCSS but we can pull this as vanilla CSS and do that later, along with integrating the linter. Please just make sure it conforms to the documented standards :)

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+.navbar-brand > .icon-banner {
+  position: relative;
+  top: -2px;
+  display: inline;
+}
+.icon {
+  position: relative;
+  top: -10px;
+}
+.logged-out iframe {
```
mattjbrent: Not sure what context this is being used in but we use modifiers for this kind of language, with the `is-` prefix.

This should also probably be scoped with another class otherwise this rule will apply to all iframe elements as `is-logged-out` will probably be common. Depending on the context, this should be `.my-component.is-logged-out iframe`

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+  height: 0;
+}
+.page-consent .client-logo {
+  float: left;
+}
+.page-consent .client-logo img {
+  width: 80px;
+  height: 80px;
+}
+.page-consent .consent-buttons {
```
mattjbrent: THis is a bit ambiguous without looking at the markup but something like `.page-consent--logo` or something like that is more preferable.

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+  margin-top: 25px;
+}
+.page-consent .consent-form .consent-scopecheck {
```
mattjbrent: this should probably be `.page-consent--form .consent-scopeheck`

src/Auth/wwwroot/css/site.css
mattjbrent: The reason why we use the {component}--{child-component} selector is because we can avoid specific selectors like this. By using `.page-consent--logo` we can avoid stringing two classes together and increasing the selector specificity.

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+}
+.page-consent .consent-form .consent-description {
+  margin-left: 25px;
+}
+.page-consent .consent-form .consent-description label {
+  font-weight: normal;
+}
+.page-consent .consent-form .consent-remember {
+  padding-left: 16px;
+}
```
mattjbrent: replace `.page-consent .consent-form` with `.page-consent--form`

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+.grants .page-header {
+  margin-bottom: 10px;
+}
+.grants .grant {
```
mattjbrent: `.grant--container .grant` or just `.grant` preferably just grant if you can avoid the second class

src/Auth/wwwroot/css/site.less
mattjbrent: @misha130 We should definitely be using postCSS here. Take the config from the landing page.

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+}
+.grants .grant .granttype {
+  font-size: 120%;
+  font-weight: bold;
+}
+.grants .grant .created {
+  font-size: 120%;
+  font-weight: bold;
+}
+.grants .grant .expires {
```
mattjbrent: Again, try to reduce selector specificity if you can for all of the grant selectors.

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+}
+.grants .grant img {
+  width: 100px;
+  height: 100px;
+}
+.grants .grant .clientname {
+  font-size: 140%;
+  font-weight: bold;
+}
+.grants .grant .granttype {
```
mattjbrent: `.grant__type`

src/Auth/wwwroot/css/site.css
```diff
@@ -0,0 +1,82 @@
...
+.grants .grant {
+  margin-top: 20px;
+  padding-bottom: 20px;
+  border-bottom: 1px solid lightgray;
+}
+.grants .grant img {
+  width: 100px;
+  height: 100px;
+}
+.grants .grant .clientname {
```
mattjbrent: `.grant__clientname`

src/Auth/wwwroot/js/site.js
```diff
@@ -0,0 +1,25 @@
...
+        // Position header slide appropriately relative to
+        // trigger.
+        const rect = this.getBoundingClientRect();
+        hs.style.top = (rect.top + rect.height) + "px";
+        hs.style.right = (document.body.clientWidth - rect.right) + "px";
+
+        // Prevent navigation
+        e.preventDefault();
+    });
+}
```
mattjbrent: add newline at EOF

src/Auth/wwwroot/js/site.js
```diff
@@ -0,0 +1,25 @@
+﻿/*
+ * Header slides
+ ----------------
+ * This code powers the "header slides", which are at the core of mobile nav
+*/
+
+const headerSlideTriggers = document.querySelectorAll("[data-trigger-header-slide]");
```
mattjbrent: if we arent using Babel, dont use `const`. Replace with `var`

src/Auth/wwwroot/js/site.js
```diff
@@ -0,0 +1,25 @@
...
+
+for (let i = 0; i < headerSlideTriggers.length; i++) {
```
mattjbrent: same as above, replace `let` with `var`

src/Auth/wwwroot/js/site.js
```diff
@@ -0,0 +1,25 @@
...
+    headerSlideTriggers[i].addEventListener("click", function (e) {
+        const headerSlide = document.querySelector(this.getAttribute("data-trigger-header-slide"));
```
mattjbrent: `const` -> `var`

src/Auth/wwwroot/js/site.js
```diff
@@ -0,0 +1,25 @@
...
+
+        headerSlide.classList.toggle("is-active");
+        this.classList.toggle("is-active");
+
+        // Position header slide appropriately relative to
+        // trigger.
+        const rect = this.getBoundingClientRect();
```
mattjbrent: `const` -> `var`

mattjbrent (dismissed):

src/Auth/Models/ErrorViewModel.cs
ranolfi: Marking as resolved but feel free to bring up any additional concerns on Discord or in a new issue.

src/Auth/wwwroot/css/site.css
misha130: Removed it

src/Auth/wwwroot/css/site.css
misha130: There is bootstrap there right now. This will be removed later. I left it as is for now

src/Auth/wwwroot/css/site.css
misha130: I checked, both are ok

src/Auth/Views/Account/Logout.cshtml
ranolfi: We must obtain the client name somehow then...

luap42: Is it possible to access either the settings db or the settings file there?

ArtOfCode-: I'd like this to show "Would you like to log out of $SiteName" - doesn't matter where that comes from, but presumably it'll be a configurable item so that folks running their own can rename the site. Can we access those site settings here?

ranolfi: Yes, I believe this should come from a method or property defined in the Infrastructure project.

src/Auth/SeedDatabase.cs
misha130: Ended up adding a generate password method in https://github.com/codidact/core/pull/29/commits/285787fc53560d31c9fc5c407580d3f45d1f0d4c because GUID passwords werent up to standards of the framework.

src/Auth/Views/Account/Logout.cshtml
misha130: I did this: https://github.com/misha130/core/commit/b8be380a9481ec02c947881b60e67153eff7ea59
Its on another branch cause I don't like it much.
What do you think

src/WebUI/Views/Shared/_Layout.cshtml
misha130: I deep tested this. I switched the two lines. Checked that after I login there is "IsAuthenticated" true.  Saw that the buttons show up correctly. 

A bit weird but both cases just worked (IsAuthenticaed showed sign in, !IsAuthenticated showed up sign in)

Anyway I don't see a problem after https://github.com/codidact/core/pull/29/commits/47e2c020e2c85c1b7a0ba7d6e9c5f50277bb6c2b

src/Auth/Views/Account/Logout.cshtml
ArtOfCode-: Looks good to me, @misha130

misha130: Then I merged. I think thats it.

src/Auth/Common/SecurityHeadersAttribute.cs
```diff
@@ -0,0 +1,55 @@
...
+                    context.HttpContext.Response.Headers.Add("Content-Security-Policy", csp);
+                }
+                // and once again for IE
+                if (!context.HttpContext.Response.Headers.ContainsKey("X-Content-Security-Policy"))
+                {
+                    context.HttpContext.Response.Headers.Add("X-Content-Security-Policy", csp);
+                }
+
+                // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy
+                var referrer_policy = "no-referrer";
```
ArtOfCode-: Should be `referrerPolicy`, or could just be inlined altogether.

src/Auth/SeedDatabase.cs
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        /// <summary>
+        /// Generates a Random Password
+        /// respecting the given strength requirements.
+        /// </summary>
+        /// <param name="opts">A valid PasswordOptions object
+        /// containing the password strength requirements.</param>
+        /// <returns>A random password</returns>
+        private static string GenerateRandomPassword()
```
ArtOfCode-: **This function is a security concern.**

 * First and foremost - if C# provides a native function to generate a random password, use that instead. Never roll your own security when tested implementations are available. [`System.Web.Security.Membership.GeneratePassword`](https://docs.microsoft.com/en-us/dotnet/api/system.web.security.membership.generatepassword?redirectedfrom=MSDN&view=netframework-4.8#System_Web_Security_Membership_GeneratePassword_System_Int32_System_Int32_) may fit that description. Failing that, the following will make this more secure:

 * Using `Environment.TickCount` to seed a `Random` instance may be acceptable for non-security use-cases, but particularly in password generation it represents a vulnerability because it's predictable.

 * Security-conscious functions should use a secure random number generator; `System.Random` is a PRNG, not a csPRNG. Use `System.Security.Cryptography.RNGCryptoServiceProvider` or equivalent instead:

    ```c#
    // Put provider somewhere accessible to everything that needs random values.
    var provider = new System.Security.Cryptography.RNGCryptoServiceProvider();

    var randomBytes = new byte[4];
    provider.GetBytes(randomBytes);

    var randomInteger = BitConverter.ToUInt32(randomBytes, 0);
    ```

 * The way this method is written inserts required character types in a predictable order, which reduces the effective entropy of the resulting password. Instead of checking whether a character type is required and inserting one if so, try something like this:

    ```c#
    string[] randomChars = new[] { ... }; // as you had it already
    string availableChars = "";
    
    if (opts.RequireUppercase)
    {
        availableChars += randomChars[0];
    }
    
    if (opts.RequireLowercase) 
    {
        availableChars += randomChars[1];
    }
    
    // etc
    // continue with your for loop unchanged
    ```

ArtOfCode- (changes requested):

src/Auth/Common/SecurityHeadersAttribute.cs
misha130: https://github.com/codidact/core/pull/29/commits/3ee072818d71c72449a48a61e422125a8e32fd60

src/Auth/SeedDatabase.cs
misha130: > First and foremost - if C# provides a native function to generate a random password, use that instead. Never roll your own security when tested implementations are available. System.Web.Security.Membership.GeneratePassword may fit that description.

.net core doesn't. Generating passwords is a security flaw because 
>  ...password generation it represents a vulnerability because it's predictable

Maybe just open user creation page if there is no user on launch? And I'll delete the whole class?

Or could use https://github.com/prjseal/PasswordGenerator/ but it also doesn't follow the above standards https://github.com/prjseal/PasswordGenerator/blob/077b798a3a86679e70171a5959db2c962dfbd912/PasswordGenerator/Password.cs#L183

luap42: > Maybe just open user creation page if there is no user on launch?

Not sure, whether that is good for implementation, but from a user point of view running a "setup wizard" when there is no user seems like a good idea. That one could configure:

- The first user account
- The default site settings (i.e. name)
- ...

ArtOfCode-: That's the ideal way to do it, yes - I originally suggested just generating a password and printing it because that seemed to be the least amount of work.

misha130: I enabled the user to input his data via console. Password is invisible https://github.com/codidact/core/pull/29/commits/58e189dd03a7f1d66eb32fc781e696c34d78a6f6

ArtOfCode- (approved):

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+        [HttpGet]
+        public IActionResult AccessDenied()
+        {
+            return View();
+        }
+
+
+        /*****************************************/
+        /* helper APIs for the AccountController */
+        /*****************************************/
```
asynts: This violates the coding standards.

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+                    // this will send back an access denied OIDC error response to the client.
+                    await _interaction.GrantConsentAsync(context, ConsentResponse.Denied);
+
+                    // we can trust model.ReturnUrl since GetAuthorizationContextAsync returned non-null
+                    if (await _clientStore.IsPkceClientAsync(context.ClientId))
+                    {
+                        // if the client is PKCE then we assume it's native, so this change in how to
+                        // return the response is for better UX for the end user.
+                        return View("Redirect", new RedirectViewModel { RedirectUrl = model.ReturnUrl });
+                    }
```
asynts, line 86: `Redirect(model.ReturnUrl)` should be enough.

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+                    await _events.RaiseAsync(new UserLoginSuccessEvent(user.UserName, user.Id.ToString(), user.UserName, clientId: context?.ClientId));
+
+                    if (context != null)
+                    {
+                        if (await _clientStore.IsPkceClientAsync(context.ClientId))
+                        {
+                            // if the client is PKCE then we assume it's native, so this change in how to
+                            // return the response is for better UX for the end user.
+                            return View("Redirect", new RedirectViewModel { RedirectUrl = model.ReturnUrl });
+                        }
```
asynts, line 112: *ditto*.

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+                        return Redirect(model.ReturnUrl);
+                    }
+                    else if (string.IsNullOrEmpty(model.ReturnUrl))
+                    {
+                        return Redirect("~/");
+                    }
+                    else
+                    {
+                        // user might have clicked on a malicious link - should be logged
+                        throw new Exception("invalid return URL");
```
asynts: **That's not "logging"!**

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            // something went wrong, show form with error
+            var vm = await BuildLoginViewModelAsync(model);
+            return View(vm);
+        }
+
+
+        /// <summary>
+        /// Show logout page
+        /// </summary>
```
asynts, line 146: This comment does not relay any information.

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+        [HttpGet]
+        public async Task<IActionResult> Logout(string logoutId)
+        {
+            // build a model so the logout page knows what to display
```
asynts, line 150: *ditto*.

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+        }
+
+        /// <summary>
+        /// Handle logout page postback
+        /// </summary>
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<IActionResult> Logout(LogoutInputModel model)
+        {
+            // build a model so the logged out page knows what to display
```
asynts, line 170: *ditto*.

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+            }
+
+            return View("LoggedOut", vm);
+        }
+
+        [HttpGet]
+        public IActionResult AccessDenied()
+        {
+            return View();
+        }
```
asynts, line 206: There should be an universal error page, for example `/error/403/access-denied`, that would fit the URL schema.

*Note. This an endpoint form the Identity framework; the framework can be configured to target a different endpoint.*

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+
+            var vm = new LoggedOutViewModel
+            {
+                AutomaticRedirectAfterSignOut = AccountOptions.AutomaticRedirectAfterSignOut,
+                PostLogoutRedirectUri = logout?.PostLogoutRedirectUri,
+                ClientName = string.IsNullOrEmpty(logout?.ClientName) ? logout?.ClientId : logout?.ClientName,
+                LogoutId = logoutId
+            };
+
+            if (User?.Identity.IsAuthenticated == true)
```
asynts:
```suggestion
            if (User?.Identity.IsAuthenticated)
```

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// Handle logout page postback
+        /// </summary>
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        public async Task<IActionResult> Logout(LogoutInputModel model)
+        {
+            // build a model so the logged out page knows what to display
+            var vm = await BuildLoggedOutViewModelAsync(model.LogoutId);
+
+            if (User?.Identity.IsAuthenticated == true)
```
asynts, line 173:
```suggestion
            if (User?.Identity.IsAuthenticated)
```

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+            vm.Username = model.Username;
+            vm.RememberLogin = model.RememberLogin;
+            return vm;
+        }
+
+        private async Task<LogoutViewModel> BuildLogoutViewModelAsync(string logoutId)
+        {
+            var vm = new LogoutViewModel { LogoutId = logoutId, ShowLogoutPrompt = AccountOptions.ShowLogoutPrompt };
+
+            if (User?.Identity.IsAuthenticated != true)
```
asynts:
```suggestion
            if (!User?.Identity.IsAuthenticated)
```

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+            {
+                // if the user is not authenticated, then just show logged out page
+                vm.ShowLogoutPrompt = false;
+                return vm;
+            }
+
+            var context = await _interaction.GetLogoutContextAsync(logoutId);
+            if (context?.ShowSignoutPrompt == false)
```
asynts:
```suggestion
            if (!context?.ShowSignoutPrompt)
```

src/Auth/Views/Shared/_ValidationSummary.cshtml
```diff
@@ -0,0 +1,7 @@
+﻿@if (ViewContext.ModelState.IsValid == false)
```
asynts, line 1:
```suggestion
﻿@if (!ViewContext.ModelState.IsValid)
```

src/Auth/Controllers/AccountController.cs
```diff
@@ -0,0 +1,301 @@
...
+        /// <summary>
+        /// Show logout page
+        /// </summary>
+        [HttpGet]
+        public async Task<IActionResult> Logout(string logoutId)
+        {
+            // build a model so the logout page knows what to display
+            var vm = await BuildLogoutViewModelAsync(logoutId);
+
+            if (vm.ShowLogoutPrompt == false)
```
asynts, line 153:
```suggestion
            if (!vm.ShowLogoutPrompt)
```

src/Auth/Views/Account/LoggedOut.cshtml
```diff
@@ -0,0 +1,32 @@
...
+    }
+</div>
+
+@section scripts
+{
+    @if (Model.AutomaticRedirectAfterSignOut)
+    {
+        <script src="~/js/signout-redirect.js"></script>
+    }
+}
```
asynts: Why do the redirect on the client side if that information is known on the server side? The implementation is also very messy because it looks at the DOM and extracts the return URL from there.

src/Auth/Models/LoginViewModel.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace Codidact.Auth.Models
+{
+    public class LoginViewModel : LoginInputModel
+    {
+        public bool AllowRememberLogin { get; set; } = true;
+    }
```
asynts, line 9: The classes `LogoutViewModel` and `LogoutInputModel` could be merged.

src/Auth/Models/LogoutViewModel.cs
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Brock Allen & Dominick Baier. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See LICENSE in the project root for license information.
+
+namespace Codidact.Auth.Models
+{
+    public class LogoutViewModel : LogoutInputModel
+    {
+        public bool ShowLogoutPrompt { get; set; } = true;
+    }
```
asynts, line 9: The classes `LogoutViewModel` and `LogoutInputModel` could be merged.

src/Auth/Program.cs
```diff
@@ -0,0 +1,39 @@
...
+using Serilog.Sinks.SystemConsole.Themes;
+using System;
+
+namespace Codidact.Auth
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            Console.Title = "Codidact Auth";
```
asynts, line 17: That's a bit short, how about "Codidact Authentication Server".

src/Auth/Properties/launchSettings.json
```diff
@@ -0,0 +1,11 @@
+{
+  "profiles": {
+    "Auth": {
+      "commandName": "Project",
+      "environmentVariables": {
+        "ASPNETCORE_ENVIRONMENT": "Development"
+      },
+      "applicationUrl": "http://localhost:5000"
```
asynts:
```suggestion
      "applicationUrl": "https://localhost:5000"
```

There are a few other places where this needs to be changed.

src/Auth/SeedDatabase.cs
```diff
@@ -0,0 +1,73 @@
...
+
+namespace Codidact.Auth
+{
+    public static class SeedDatabase
+    {
+        /// <summary>
+        /// Initializes identity data like users.
+        /// </summary>
+        /// <param name="app"></param>
+        /// <returns></returns>
```
asynts, line 17: This comment relays no information.

src/Auth/SeedDatabase.cs
```diff
@@ -0,0 +1,73 @@
...
+            else
+            {
+                Console.WriteLine($"Error Creating user:");
+                foreach (var error in result.Errors)
+                {
+                    Console.WriteLine($"({error.Code}) {error.Description}");
+                }
+                await CreateInitialUser(userManager);
+            }
+        }
```
asynts, line 71: This could be loaded from the configuration instead.

src/Auth/appsettings.json
```diff
@@ -0,0 +1,42 @@
...
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft": "Warning",
+      "Microsoft.Hosting.Lifetime": "Information"
+    }
+  },
+  "AllowedHosts": "*",
+  "IdentityServer": {
+    "IssuerUri": "urn:sso.company.com",
```
asynts: That looks like a placeholder.

src/Auth/appsettings.json
```diff
@@ -0,0 +1,42 @@
...
+    "Clients": [
+      {
+        "Enabled": true,
+        "ClientId": "codidact_client",
+        "ClientName": "Codidat Client",
+        "ClientSecrets": [ { "Value": "SXPCSlOE2D+8lXHF2+cqr+oD5UpMdrhkMeMOCJLJHl4=" } ],
+        "AllowedGrantTypes": [ "authorization_code" ],
+        "AllowedScopes": [ "openid", "profile" ],
+        "RedirectUris": [ "http://localhost:5001/signin-oidc" ],
```
asynts:
```suggestion
        "RedirectUris": [ "https://localhost:5001/signin-oidc" ],
```

asynts (commented):
*Note. My laptop gave up on me before I could really complete this review (the arrow keys aren't working anymore for some reason.) I can't easily verify some of the suggested changes, these should be tested carefully before being accepted.*

There are two major things that must be addressed in my opinion:

1. Many source files have a copyright notice in them, referencing the contents of the `LICENSE` file. This file contains Codidact's license, not the license for these files.

 2. I found the following code in the `AccountController.cs` file:
    
        // user might have clicked on a malicious link - should be logged
        throw new Exception("invalid return URL");

    This essentially means that the authentication server will crash if an invalid return URL is provided. ~~It is true that it's not supposed to be possible for that to happen (because the CSRF token is verified),  however, a big red error message in the logs should suffice, in my opinion.~~(no, the application actually crashes if a user accesses `http://localhost:5000/identity/account/login?returnurl=foo` and logs in.)  A log message should be enough, info level should suffice.

src/Auth/Controllers/AccountController.cs
misha130: You are suggesting I remove the whole `if` because a line later it says `Redirect(model.ReturnUrl)`.

This JavaScript is better in case the one authenticating is a native mobile application its not exactly as simple as redirect:
`<meta http-equiv="refresh" content="0;url=@Model.RedirectUrl" data-url="@Model.RedirectUrl">
`

I have to be honest, I am not very sure how this is different and how it works for native mobile apps but I want to trust IdentityServer that they knew what they were doing

src/Auth/Controllers/AccountController.cs
misha130: (Note I accidently deleted this file https://github.com/IdentityServer/IdentityServer4/blob/76dfa41b9648b064b43552190491b917fda40193/samples/Quickstarts/5_EntityFramework/src/IdentityServer/Views/Account/AccessDenied.cshtml so this is a bug on its own)


This one I tried to write something like this:

```
public IActionResult AccessDenied()
        {
            return View("Error", new ErrorViewModel
            {
                Error = new ErrorMessage
                {
                    Error = "Access Denied",
                    ErrorDescription = "You do not have access to this resource.",
                },
            });
        }
```

But then I realized this page doesn't really fit it https://github.com/codidact/core/blob/58e189dd03a7f1d66eb32fc781e696c34d78a6f6/src/Auth/Views/Shared/Error.cshtml

What do you think? Should I bring back the original page or use the generic error?

src/Auth/Views/Account/LoggedOut.cshtml
misha130: I tried removing this but I had to do extensive testing and discovered the automatic sign out didn't work because callback sign out URL never came in. 
In the end I managed to remove this but I really hope that this is not some special requirement for some special situation that IdentityServer opts in for. It doesn't seem so at least.
https://github.com/misha130/core/commit/beda0a29086e2018e450f49a57265615330202ca

src/Auth/SeedDatabase.cs
misha130: It was initially loaded from configuration.
But then @ArtOfCode- suggested we should generate a password.
But then all generated passwords are deemed unsecure.
So this is the end result of that, you can see these discussions in this PR.

src/Auth/Controllers/AccountController.cs
asynts: My original suggestion was to tell the Identity framework to go to `/error/403/access-denied` thus leaving the authentication server completely. On a second thought that might not be a good idea after all, mainly because it wouldn't be possible to relay any additional details if there were some.

From the alternatives you proposed, a static error page should be enough. I don't see many other errors that could happen during authentication (a 404 and such aren't handled by the authentication server anyways.)

misha130: Fixed by https://github.com/codidact/core/pull/29/commits/3e8ca2f42ee1f30799f2ac11339795f2f5e95e77

src/Auth/Controllers/AccountController.cs
asynts: Personally, the code shouldn't exist if we can't explain why it's needed. I would suggest removing it and if it breaks later we can re-add the code with a comment explaining why it's needed.

misha130: Well, I went and checked it more.
As I understood it this moving from page to page is part of https://tools.ietf.org/html/rfc7636
The 302 that Redirect does is incorrect and is unsafe for injections.
Also extra reading: https://christianlydemann.com/implicit-flow-vs-code-flow-with-pkce/

src/Auth/SeedDatabase.cs
asynts: *Note. I caught up with all the comments (that I found).*

I would suggest to simply hard-code an admin with password for MVP. I think that's how it originally was anyway.

Then post-MVP we could add a visual setup wizard, this seems to be standard for such web applications. Then the first user visiting the applications would be able to configure all of this in his web-browser.

misha130:
> This essentially means that the authentication server will crash if an invalid return URL is provided. It is true that it's not supposed to be possible for that to happen (because the CSRF token is verified), however, a big red error message in the logs should suffice, in my opinion.(no, the application actually crashes if a user accesses http://localhost:5000/identity/account/login?returnurl=foo and logs in.) A log message should be enough, info level should suffice.

The server doesn't crash but you get a whole error page.
https://i.gyazo.com/f0f1dd94a4c1296a579c408ced1ab2b6.png

Please note this is a POST not a GET. A POST after pushing the Login button.
If this happens this is 100% malicious. An exception like this is very appropriate.
All exceptions should be logged (future dev.). So adding a `Log.Error` line here is redundant

src/Auth/SeedDatabase.cs
ArtOfCode-: No need to change this further now. This is fine as it is, @misha130 - just waiting on review from @ranolfi to be able to merge.

ranolfi: The (still unofficial?) specs include a separate application service for authentication. Should we merge this now and refactor later (to a separate repo/project)? @misha130, @DoctaJonez

src/Auth/Properties/launchSettings.json
misha130: I made more complicated ports but made it work: 3aa6174d

src/Auth/SeedDatabase.cs
misha130: @ranolfi no, rather create a rep and move it by hand

src/Auth/appsettings.json
misha130: 3aa6174d

src/Auth/appsettings.json
misha130: 3d3ef2a9

src/Auth/Controllers/AccountController.cs
misha130: Exceptions should be logged. No need to actually add another log here

src/Auth/Controllers/AccountController.cs
misha130: Moved it to another service: https://github.com/codidact/core/pull/29/commits/356b742ceec438731a0fcfebda78082f8c1800a8

src/Auth/SeedDatabase.cs
ranolfi: @asynts, this is auto-generated [XML documentation](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/xmldoc/). It shows up in Intellisense and can be used to auto-generate docs in a build.

asynts:
> So adding a `Log.Error` line here is redundant

I didn't describe my suggestion particularly well, what I meant was to create a log message instead of throwing an exception. The request could then be terminated by responding with "400 Bad Request".

  > If this happens this is 100% malicious.

@misha130 I agree. However, the exception doesn't carry any information about the malicious URL. It seems like a "lazy" way to deal with an edge case and could be replaced with something like `return Redirect("/error/400/bad-request");`. This would be a "more correct" response in my opinion.

If the intention is to log such a request, this could be done using `throw new Exception($"invalid return url '{model.ReturnUrl}'");` or better, `_logger.LogInformation($"invalid return url '{model.ReturnUrl}'");`.

src/Auth/Controllers/AccountController.cs
asynts: > Since the access token is sent as the fragment (hash) part of the redirect URL (also called the front channel), it will be exposed to an attacker if the redirect is intercepted. Now, of course, it is recommended to always use HTTPS, but even HTTPS can be compromised by man-in-the-middle attacks with methods as corporate SSL certificates to inspect traffic.
  >
  > This threat is mitigated in code flow by not getting the token response from a redirect but instead a direct request-response HTTP request.

I find this questionable because if redirect are compromised, why not a normal request too? I wasn't able to find anything else that supports that.

However, most of these stories have some sort of truth to them. Personally, the best action would be to keep the code and add a comment to that article.

misha130:
> and could be replaced with something like return Redirect("/error/400/bad-request");

This is not the place to write this though. It should be configured server wide.
Plus if this is malicious I want the user to have as little information as possible.

I did this as per your suggestion and added more information to the exception:
`throw new Exception($"Invalid return url '{model.ReturnUrl}'");`

Beyond this I'd rather not amend it

src/Auth/Controllers/AccountController.cs
asynts: I think I found the problem this is based on:

  > ## An important note about referers on older browsers
  > 
  > Even though HTTPS encrypts the path you are accessing, if you click a
  > hyperlink within that site which goes to an unencrypted page, the full path
  > may be leaked in the `referer` header. This is [not the case anymore][1] for
  > many newer browsers, but older or non-compliant browsers may still have this
  > behavior, as will websites which set the HTML5 referer meta tag to always
  > send the information. An example sent *unencrypted* by a client to jump from
  > `https://example.com/private/details.html` to
  > `http://example.org/public/page.html` in such a case would be:
  > 
  > ~~~none
  > GET /public/page.html
  > Referer: https://example.com/private/details.html
  > User-Agent: Wget/1.19.1 (linux-gnu)
  > Accept: */*
  > Accept-Encoding: identity
  > Host: example.org
  > ~~~ 
  > 
  > As such, navigating from an HTTPS page to an HTTP page may leak the full URL
  > (excluding the fragment) of the previous page, so keep that in mind.
  > <sub>[(source)][2]</sub>

However, even this source says that 

  > As such, navigating from an HTTPS page to an HTTP page may leak the full URL
  > **(excluding the fragment)** of the previous page, so keep that in mind.
  > <sub>[(source)][2]</sub>

which contradicts

  > Since **the access token is sent as the fragment** (hash) part of the redirect URL (also called
  > the front channel), it will be exposed to an attacker if the redirect is intercepted.
  > <sub>[(source)][3]</sub>

  [1]: http://smerity.com/articles/2013/where_did_all_the_http_referrers_go.html
  [2]: https://security.stackexchange.com/a/176169/223751
  [3]: https://christianlydemann.com/implicit-flow-vs-code-flow-with-pkce/

ranolfi (approved):

ranolfi:
Excellent work, everyone.

Right now I'm considering whether to merge this and then refactor to the new (soon-to-be-created) 'authentication-service' repo, or close it and re-submit as intended. Leaning towards the former.

ranolfi:
This is a webhook test.

src/Auth/Controllers/AccountController.cs
asynts: I figured it out, [it's for user experience][1].

  [1]: https://github.com/IdentityServer/IdentityServer4.Demo/blob/1f98c8012d7ff08c9a198660f3fe6ad8dafc6002/src/IdentityServer4Demo/Quickstart/Account/AccountController.cs#L95-L96

misha130: Alright. We could technically remove it but I guess better UX is ok

Issue #36, opened 2020-01-28
Privacy policy page is empty


jflopezfernandez:
The privacy policy page is currently empty. I'm currently working on populating the page with [the privacy policy from the forum](https://forum.codidact.org/privacy).

ArtOfCode-:
Note to all contributors: See linked pull request #37. We'll need to write a custom policy when we're further along in development; please hold off on implementing this for now.

asynts:
I think this could be closed, it will be quite some time until we add the privacy policy. Alternative this could be added to a milestone, e.g. "final sprint".

luap42:
Closing this with three different status tags, because this has been declined (see the other PR), because we'll need more information on how Codidact will collect data (because we can't know yet), and the idea behind implementing this now is therefore invalid. :)

PR #37, opened 2020-01-28
Jflopezfernandez/36/populate privacy policy page

jflopezfernandez:
## Fixes Issue #36 

This commit populates the site privacy policy page with the privacy policy from [the forum](https://forum.codidact.org/privacy). No modifications were made to the privacy policy, I simply copied and pasted it, and then added basic html markup to each element.

Updated Privacy Policy Page

### DIFF
--- a/src/WebUI/Views/Home/Privacy.cshtml
+++ b/src/WebUI/Views/Home/Privacy.cshtml
@@ -3,4 +3,65 @@
 }
 <h1>@ViewData["Title"]</h1>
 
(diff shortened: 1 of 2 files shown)


ArtOfCode-:
Gonna decline this one - thanks for thinking of it, but we'll need to write a customized privacy policy for Codidact Q&A when we're much further along in its creation, because we'll collect and use different data in different ways.

jflopezfernandez:
Yea, that makes sense, no problem.

Issue #38, opened 2020-01-28
Registration Page Needed


jflopezfernandez:
There is currently no registration page for the site. I am currently working on implementing it.

LionelBergen:
@ArtOfCode- Sounds like this is no longer in progress and we are waiting for wire frames
https://github.com/codidact/core/pull/39#issuecomment-580927220

luap42:
Agreed. Closing this for now.

cellio:
I got here from the "done" category on the project, but it sounds like this was abandoned not done.

PR #39, opened 2020-01-28
Registration Page Created

jflopezfernandez:
## Fixes Issue #38 

Created a basic registration page using the codidact co-design CSS
framework. There is nothing linking to the registration page at the moment,
so the only way to reach the registration page is via the /Home/Register URI.

I still need to make a few changes suggested by the front end team.

### DIFF
--- a/src/WebUI/Controllers/HomeController.cs
+++ b/src/WebUI/Controllers/HomeController.cs
@@ -19,6 +19,11 @@ namespace Codidact.WebUI.Controllers
             return View();
         }
 
+        public IActionResult Register()
+        {
+            return View();
+        }
+
         public IActionResult Privacy()
         {
             return View();
--- a/src/WebUI/Views/Home/Register.cshtml
+++ b/src/WebUI/Views/Home/Register.cshtml
@@ -6,7 +6,11 @@
 
 <form method="post" action="/Home/Register">
     <div>
-        <label class="form-element" id="label-email" for="input-email">Email</label>
+        <div class="notice" id="notice-email-privacy-information" style="display: none; z-index: 1;">
+            <p>Your email will be used solely for internal purposes, such as account recovery and important communication. We do not share this information with any third parties.</p>
+        </div>
+        
+        <label class="form-element" id="label-email" for="input-email">Email<button type="button" class="button is-icon-only-button" id="button-email-info" onclick="toggleViewEmailPrivacyInfo()">&#63;</button></label>
         <input type="email" class="form-element" id="input-email" name="email" autocomplete="off" required="required" aria-labelledby="label-email" />
     </div>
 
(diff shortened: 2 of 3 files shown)


jflopezfernandez:
## Fixes Issue #38 

The registration page prototype is ready for review. If the backend team can check it out and make sure I didn't miss any required information inputs or anything like that, that would be awesome. The email privacy notice is hard-coded, so I need to look into how we're planning on carrying out I18N and L10N.

Look forward to any feedback,
Jose

Registration Page Prototype

asynts:
*Note. Not sure if the following is constructive or not, I don't do front-end stuff normally.*

The CSS in this pull request is mostly done inline using `style="..."` and could be done with classes from the co-design framework instead. I tried to change this by using the co-design framework as much as possible, [here is the code][1], this is what it looks like:

Screenshot from 2020-01-29 14-10-06

I know! It doesn't look like the form from the pull request and is objectively ugly. However, it would be more coherent with the look of the remaining website.

Maybe it would be best to go with some variation of this and then adopt the co-design framework to make it look good?

For comparison, here is how it looks like currently:

Screenshot from 2020-01-29 14-03-16

  [1]: https://github.com/asynts/codidact-core/blob/asynts/38/use-codesign-classes/src/WebUI/Views/Home/Register.cshtml

jflopezfernandez:
@asynts Yea, the code looks terrible right now lol. I was prototyping as fast as possible last night while a bunch of the front-end team was on discord so I could get their feedback on what the form should look like, and then I could just refactor once I had the final design done. I also need to look at the html/css coding standards to make sure the refactored code looks alright.

As far as your redesign, it looks great to me, but I am basically illiterate when it comes to UI, that's why I was bouncing designs off the team yesterday lol

luap42:
Regarding that, I have created a JSFiddle with a similar draft to @asynt|'s one using Co-Design:

https://jsfiddle.net/neubx8wv/2/

It also looks nice with a grey background:

https://jsfiddle.net/neubx8wv/3/

And here a similar version to the original one:

https://jsfiddle.net/neubx8wv/4/

And at last without grey background:

https://jsfiddle.net/neubx8wv/5/

jflopezfernandez:
@luap42 Damn, that gray background does look nice

asynts:
> Yea, the code looks terrible right now lol.

@jflopezfernandez It seems, you misunderstood what I meant by "objectively ugly". I was referring to my own design, in my opinion it's inferior &mdash; design wise &mdash; but with cleaner code.

~~@luap42 Personally, the second fiddle looks best. I would suggest a bold title, making it stand out more:~~ Just read @mattjbrent's comment, scratch this, for now.

~~https://jsfiddle.net/asynts/cymo37j4/1/~~

jflopezfernandez:
There were a few problems with this PR, specifically:
- It was added to the wrong project, per misha
- It used MVC instead of MVVM, per Marc
- We should really have design documents and wireframes done first, per mattjbrent

I'm going to go ahead and close this pull request to clear out the queue, although we should definitely incorporate the design suggestions here. (Especially that gray background c: )

Issue #40, opened 2020-01-29
Minor change: Remove magic number from communities database migration file


jflopezfernandez:
Modify communities database migration file to use the `CommunitiyStatus` enum directly, rather than simply a value of `1`. Currently in progress.

jflopezfernandez:
I'm not too familiar with Entity Framework, and was trying to modify a migration file that cannot actually be modified, as they are auto-generated.

PR #41, opened 2020-01-29
Fixes issue #40

jflopezfernandez:
## Fixes Issue #40 
Replaced magic number `1` with the actual value from the `CommunityStatus` enum.


jflopezfernandez:
As @misha130 explained to me, migration files are not editable, so this change is essentially pointless. He did, however, mention that the change to the db file would have the desired effect. I will make the change in a different PR, and in the meantime am closing this one.

PR #49, opened 2020-01-30
Automatically collapse automatically generated files.

### LINKED ISSUE #48: Automatically Collapse Generated Files in GitHub Diffs
It's so annoying to see all the migration stuff in the diffs, apparently this can be configured in the `.gitattributes` file.

https://thoughtbot.com/blog/github-diff-supression

I'll look into that.

asynts:
Fixes #48 

I've tested this with a dummy repository, this seems to work.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/src/Infrastructure/Persistence/Migrations/* linguist-generated


ranolfi (approved):
I'm approving *without* testing as it seems extremely likely to work as expected without surprises, is easy to revert if necessary, and very low risk.

Still if someone else can indeed test before merging, that'd be great.

ArtOfCode- (approved):

PR #52, opened 2020-02-01
Use local tool for EF

### LINKED ISSUE #51: Use local tools for ef
**Is your feature request related to a problem? Please describe.**
Setup instructions require us to use the dotnet-ef tool globally. This wasn't a problem pre-3.0 since the EF version was tied to the SDK version in `global.json`. It's not anymore and different developer machines could end up with tool versions eventually.

**Describe the solution you'd like**
The tool should be installed locally for the project instead, so that there's a guarantee that every developer ends up using the same local tool version.

gldraphael:
### Pull Request summary

Adds a local tool manifest file to lock the EF version, and updates documentation to not require the tool to be globally installed.

Closes #51 

### Pull Request Tasklist

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -1,0 +1,12 @@
+{
+  "version": 1,
+  "isRoot": true,
+  "tools": {
+    "dotnet-ef": {
+      "version": "3.1.1",
+      "commands": [
+        "dotnet-ef"
+      ]
+    }
+  }
+}
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ and would like to run one locally. These instructions may change heavily dependi
     "DefaultConnection": "Server=localhost;Database=YOUR_DATABASE;"
 }
 ```
-4. Open a terminal at the Codidact solution folder. Run the command `dotnet tool install --global dotnet-ef`.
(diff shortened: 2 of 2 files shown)


ranolfi (changes requested):
Good issue and good implementation, thanks.

But can you please improve the commit messages themselves? As it currently stands, they are not exactly descriptive.

For example: "Add a local tool manifest file **for the 'dotnet-ef' dependency**" or something like that for the subject line, and then in the commit body add a short description of *why* the change is being made - pretty much what you wrote in issue #51, but in fewer words if you prefer. For the second commit make sure to specify *which* tool you're referring to, so that someone reading the commit logs can have a clear idea of what exactly they are about without having to load diffs. For further refinement, I'd suggest something like "README: update instructions for local EF Core dependency instead of global".

For guidelines, please refer to: https://github.com/codidact/core/wiki/Committing-guidelines.

gldraphael:
I was expecting this to be squash merged, so the PR description gets carried over.

ranolfi:
> I was expecting this to be squash merged, so the PR description gets carried over.

Right then, can do.

ranolfi (approved):

misha130 (approved):

Issue #26, opened 2020-01-21
Add a C# linter and configure for Microsoft style guide


ArtOfCode-:
We should have a linter in the project to make it easy for folks to automatically fix their code to the style guide.

Style guide here: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions

Whatever we choose, we'll need to configure it for those settings.

asynts:
https://github.com/dotnet/format

Not sure if it follows our exact style guide, but we could just use whatever this tool spits out instead. I mean style guides are mostly for consistency and this would work on Windows and Linux.

ranolfi:
See https://github.com/codidact/core/blob/ranolfi/skeleton/Solution/.editorconfig (not final; still pending increased severity for some rules).

misha130:
Maybe just copy it off rosalyn for example?
https://github.com/dotnet/roslyn/blob/master/.editorconfig

Your editorconfig @ranolfi forbade using lambda without a body and methods without a body

ranolfi:
> Maybe just copy it off rosalyn for example?

Bad idea IMO, see for instance: `[*.{cs,csx,vb,vbx}]`. But use as reference, definitely.

> Your editorconfig @ranolfi forbade using lambda without a body and methods without a body

Can you give an example in code for both?

misha130:
> Can you give an example in code for both?
Lambda: https://github.com/codidact/core/blob/13171a1cbdf1122c819a6eda24f936fd02a6d766/src/WebUI/Startup.cs#L54
Method: 
https://github.com/codidact/core/blob/13171a1cbdf1122c819a6eda24f936fd02a6d766/src/WebUI/Program.cs#L14

LionelBergen:
I can work on this.

ranolfi:
Oh damn, I wrote a lengthy answer here @misha130, my browser crashed and I was hoping GitHub would save a draft but it's gone :(

I was going to say that is by-design but, on second thought, that's not entirely a good thing, so I'll be comparing options. Anyway, no harm in accepting official standard right now and replacing any specific rule (along with necessary refactoring, which I step forward to do myself) if we so desire in the next days.

asynts:
@ArtOfCode- This issue can be closed.

PR #54, opened 2020-02-02
linter

LionelBergen:
Added `dotnet-format`
Ran `dotnet-format` (a few minor whitespace changes)
Added the default `.editorconfig`

Relates to: https://github.com/codidact/core/issues/26

Notes:
There is no way to check for violations without automatically fixing them using this tool. 
For example, we may want to run a check and fail before tests are run, or before a deploy. Rather than asking contributors to run it before/after a merge request.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -7,6 +7,12 @@
       "commands": [
         "dotnet-ef"
       ]
-    }
+    },
+	"dotnet-format": {
+		"version": "3.2.107702",
+		"commands": [
+			"dotnet-format"
+		]
+	}
   }
 }
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,123 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+############################### 
+# Core EditorConfig Options   # 
+############################### 
+# All files 
+[*] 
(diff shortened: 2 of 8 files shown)


appgurueu:
I don't know what MS's style guide says, but personally I can't stand statements like this:
```csharp
if (stuff)
{
     // do stuff
}
```
And would prefer
```csharp
if (stuff) {
     // do stuff
}
```
Could this be configured? Should it be configured?

ArtOfCode-:
@appgurueu We're using the Microsoft C# style guide, which is the former style - so, no, we shouldn't configure this to do something different.

ArtOfCode- (dismissed):

ranolfi (changes requested):
@LionelBergen welcome, and thanks. :)

If you can please make the following changes:

1. Separate commits
  - In the first, add the `.editorconfig` file - edit commit message because it says the file is empty which it isn't. Please also specify where you obtained it from. This is important information to be logged in the git history
  - add the `dotnet format` definition to the toolset
  - *then* run it on the codebase.

The first two commits can be combined into one, your call. The last one must be separated.

If you're stuck at some point, just ask ;)

LionelBergen:
@ranolfi I found it easier to re-make the

Issue #57, opened 2020-02-11
Refactor MVC to Razor Pages.


asynts:
The project currently uses the Model-View-Controller (MVC) design pattern. It was decided (citation needed) to use Razor Pages instead.

Please note, that the `src/Auth` project will be moved into a separate repository, therefor, this part of should be refactored later (?).

asynts:
I tried to find a place where this was discussed, but couldn't. The decision was made in November of 2019 I believe. I suspect that a lot of it happened in private messages?

ranolfi:
> Please note, that the `src/Auth` project will be moved into a separate repository, therefor, this part of should be refactored later (?).

Sounds like a plan to me.

> I suspect that a lot of it happened in private messages?

Some of it has, @asynts. The question "why razor pages, and not the mvc pattern?" is [pending](https://discordapp.com/channels/634104110131445811/637322644718616576/674478787928391701) of documentation in the [developer FAQ](https://github.com/codidact/docs/tree/master/Developer-FAQ).  
If one digs deep enough on the old #programming channel (now #programming-general) and in #tech-lead-comms, something useful _may_ come up should dev-docs require it. But other than that, I'll quote what I have easily at hand.

DM with @kieranirving (frontend group leader):

> >     Why Razor pages over MVC pattern
> 
> This was discussed mainly via voice chat, but you can find some references in this post https://discordapp.com/channels/634104110131445811/641360136908046336/668175133055254528 and see the conversation/context.
> 
> The short answer is simple: because it fits better with the general architecture we're going with, in terms of reduced complexity.

Many online resources cover Razor Pages vs MVC in ASP.NET \[Core\]. Two of the most relevant IMO are present in that message from the Discord link above: 

- https://hackernoon.com/asp-net-core-razor-pages-vs-mvc-which-will-create-better-web-apps-in-2018-bd137ae0acaahttps://hackernoon.com/asp-net-core-razor-pages-vs-mvc-which-will-create-better-web-apps-in-2018-bd137ae0acaa
- https://stackify.com/asp-net-razor-pages-vs-mvc/

Additionally:

- https://stackoverflow.com/questions/46777404/why-is-razor-pages-the-recommended-approach-to-create-a-web-ui-in-asp-net-core-2
- https://stackoverflow.com/questions/48121928/asp-net-core-2-0-razor-pages-vs-full-mvc-core

TL;DR: reduced complexity means easier maintainance as the project grows, which is one of our core long-term goals.

PR #62, opened 2020-02-14
Switch from Controllers to Razor Pages

misha130:
### Pull Request summary

This PR removes all controllers usage and switches it to Razor Pages.

The Login Page is an empty redirect to call an authorization.
The Logout Page is an empty redirect to go to the signout.

https://github.com/codidact/core/issues/57

### Pull Request Tasklist


DoctaJonez (approved):

src/WebUI/Pages/Shared/_Layout.cshtml
```diff
@@ -58,7 +58,7 @@
 
     <footer class="has-padding-4 has-border-top-style-solid has-border-top-width-1 has-border-color-tertiary-050">
         <div class="container">
-            &copy; 2020 - Codidact - <a asp-area="" asp-controller="Home" asp-action="Privacy">Privacy</a>
+            &copy; 2020 - Codidact - <a asp-page="Privacy">Privacy</a>
```
asynts:
```suggestion
            &copy; 2020 - Codidact - <a asp-page="/Privacy">Privacy</a>
```

src/WebUI/Pages/Shared/_Layout.cshtml
```diff
@@ -11,7 +11,7 @@
     <header class="header">
         <div class="container header--container">
             <div class="header--brand">
-                <a class="header--site-name" asp-area="" asp-controller="Home" asp-action="Index">Codidact</a>
+                <a class="header--site-name" asp-area="" asp-page="/Home/Index">Codidact</a>
```
asynts:
```suggestion
                <a class="header--site-name" asp-area="" asp-page="/Index">Codidact</a>
```

src/WebUI/Pages/Logout.cshtml.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class LogoutModel : PageModel
+    {
+        private readonly ILogger<IndexModel> _logger;
+
+        public LogoutModel(ILogger<IndexModel> logger)
+        {
+            _logger = logger;
+        }
+
+        public IActionResult OnGet()
```
asynts:
```suggestion
        public IActionResult OnPost()
```

I wasn't able to verify this because the authentication server didn't want to startup, but this looks like a CSRF vulnerability. Razor Pages will automatically verify CRSF tokens but that only happens for post requests.

asynts (changes requested):

src/WebUI/Pages/Logout.cshtml.cs
misha130: https://docs.microsoft.com/en-us/aspnet/core/razor-pages/?view=aspnetcore-3.1&tabs=visual-studio#xsrf
https://security.stackexchange.com/a/115810
Seems fine to me

asynts: > GET endpoints should be idempotent - **actions that enact a change and do not simply retrieve data should require sending a POST (or other HTTP method) request**. POST endpoints should not interchangeably accept GET requests with parameters in the query string. &mdash; [MDN: HTTP cookies - Cross-site request forgery (CSRF)][1]

A malicious site could include some Java Script in the background that logs you out of `codidact.com`. This issue can be resolved by validating an CSRF token, which Razor Pages will do automatically, but not for 'GET' requests. 

*Addendum.* I don't want to overplay this, this isn't a major issue, but should be fixed anyways.

  [1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Cross-site_request_forgery_CSRF

masonwheeler: @asynts It is a somewhat major issue, because [it doesn't take a malicious user](https://thedailywtf.com/articles/The_Spider_of_Doom) to trigger changes with GET requests, and even if that won't happen in this specific instance, if we don't consciously make this a part of our coding practice, it'll trip us up elsewhere.

misha130: The GET doesnt change state, it only creates a url to go to the auth instance.
Also take a look at the example by IS, they didn't add anti forgery validation:
https://github.com/IdentityServer/IdentityServer4/blob/master/samples/Quickstarts/6_AspNetIdentity/src/MvcClient/Controllers/HomeController.cs

DoctaJonez: It's probably best to change this to use POST, as using GET will cause other issues too.  Namely, browsers pre-fetch URLs they think they will need to visit, so we will end up with users being "randomly" logged out if we use GET.

https://stackoverflow.com/questions/3521290/logout-get-or-post

misha130: This is not a log out call, this is a redirect call. The actual logout is in auth. 
If there is no auth instance there is no logout.

asynts: *(The two previous comments weren't visible to me because I didn't refresh the browser before submitting this comment.)*

  > it only creates a url to go to the auth instance

I tried digging through their code and this doesn't seem to be the case:

 1. [IdentityServerAuthenticationService.cs#89][2]
 2. [DefaultUserSession.cs#L210-L222][3]

It's always difficult to tell with all the dependency injection stuff, but it seems, that the logout is done by removing the authentication cookie without any request to the identity server.

From their own documentation:

  > Typically you should prompt the user for signout (meaning require a POST), otherwise an attacker could hotlink to your logout page causing the user to be automatically logged out. &mdash;&nbsp;[docs.identityserver.io - Sign-out - Removing the authentication cookie][1]

Seems like an oversight to me.

  [1]: http://docs.identityserver.io/en/latest/topics/signout.html#removing-the-authentication-cookie
  [2]: https://github.com/IdentityServer/IdentityServer4/blob/ef7c904733add9d5675da59d01b79f51774c1a2f/src/IdentityServer4/src/Hosting/IdentityServerAuthenticationService.cs#L88

  [3]: https://github.com/IdentityServer/IdentityServer4/blob/ef7c904733add9d5675da59d01b79f51774c1a2f/src/IdentityServer4/src/Services/Default/DefaultUserSession.cs#L210-L222

misha130: > I tried digging through their code and this doesn't seem to be the case:

Signout is part of PageModel/ControllerBase, not identity server. Its part of .net core app.
https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.mvc.controllerbase.signout?view=aspnetcore-3.1

An example of explaination
https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/tree/master/1-WebApp-OIDC/1-6-SignOut#signout-action-of-the-accountcontroller

The actual source code after digging: 
https://github.com/dotnet/aspnetcore/blob/47a7fe9bea6092721e7122303b2f93e1471c3d15/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L172

asynts: Looks like you are right. It just redirects to the identity server stuff where the user is prompted if he wants to logout, clicking 'yes' would make a 'POST' request.

asynts (approved):

src/WebUI/Pages/Logout.cshtml.cs
poke: `Signout("cookie")` _will_ clear the application’s cookie though. Yes, the signout on OIDC will cause a redirect to the authentication provider (to sign out _there_) but before that, it will immediately sign out of the local application. So yes, a GET can be harmful here.

asynts: @poke Disregard the comment I made above, I later noticed my mistake and created codidact/authentication#5. This should be addressed after codidact/authentication#2 is merged, along with other things.

misha130: @asynts poke was saying that when the Signout("cookie") is executed it deletes the cookie before it moves to the auth project. I couldn't verify this to be honest and I looked at the source code page above too and I didn't see this happening. Frankly I am not sure and this needs further testing. To test need to just redirect and not logout and see if cookie is still there.

poke: That is correct. A GET on an action that will call `Signout("cookie")` will _immediately_ clear the application cookie in the HTTP response. If you pair that with a `Signout("oidc")`, then the _same_ HTTP response will _also_ contain a redirect to the remote authentication provider’s signout.

So the GET will have an effect, regardless of whether the user would follow that redirect or not.

If you don’t trust my word on this: the sign-out result [signs out each scheme](https://github.com/dotnet/aspnetcore/blob/v3.1.2/src/Mvc/Mvc.Core/src/SignOutResult.cs#L107-L110) which will eventually call [`SignOutAsync` on each authentication handler](https://github.com/dotnet/aspnetcore/blob/v3.1.2/src/Http/Authentication.Core/src/AuthenticationService.cs#L225). While the OIDC scheme will create [a redirect](https://github.com/dotnet/aspnetcore/blob/v3.1.2/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectHandler.cs#L243-L249), the cookie scheme [will also clear the cookie](https://github.com/dotnet/aspnetcore/blob/v3.1.2/src/Security/Authentication/Cookies/src/CookieAuthenticationHandler.cs#L363-L366). And that happens both in the same response.

luap42: Can we complete this discussion in order to be able to finish reviewing/merge this PR soon?

asynts: I would not rush this, it should be carefully verified and fixed. This would mean setting up a testing environment, maybe we can move it to a different issue and merge this now?

asynts: I have created #67 for this, consider this "resolved" and just merge it for now.

misha130: Switched to POST for logout

src/WebUI/Pages/Shared/_Layout.cshtml
```diff
@@ -24,7 +24,9 @@
                 }
                 else
                 {
-                    <a class="header--item" asp-page="/Logout">Sign Out</a>
+                    <form method="post">
+                        <button class="header--item" asp-page="/Logout">Sign Out</button>
```
poke: You want to move the `asp-page` to the form in order to make that the target action.

misha130: It worked either way but putting a type submit and the link on the form looks cleaner so OK.

ArtOfCode-:
There are conflicts in this PR that need resolving before merge.

Issue #78, opened 2020-03-24
Create a request to list all questions


misha130:
**Is your feature request related to a problem? Please describe.**
The questions page should display all relevant questions.
Thus an API for that is needed.

**Describe the solution you'd like**
Create a repository inside of application project called Questions and retrieve the necessary questions based on the filters given from the client. 
Validate the request, return errors if not valid.
Should be authorized.

misha130:
Closed by #88

PR #88, opened 2020-04-08
Questions

misha130:
### Pull Request summary

- Handles requests for a question list
- Displays a page for a question list
- Validation on said request
- Unit tests on said requests

https://github.com/codidact/core/issues/78

### Pull Request Tasklist


src/Application/Common/Contracts/IRequestHandler.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Threading.Tasks;
+
+namespace Codidact.Core.Application.Common.Contracts
+{
+    /// <summary>
+    /// Defines a handler for a request
+    /// </summary>
+    /// <typeparam name="TRequest">The type of request being handled</typeparam>
+    /// <typeparam name="TResponse">The type of response from the handler</typeparam>
+    public interface IRequestHandler<in TRequest, TResponse>
```
asynts, line 10: A request handler should be allowed to return a type that derives from `TResponse`. (I have not tested this.)

```suggestion
    public interface IRequestHandler<in TRequest, out TResponse>
```

src/Domain/Entities/Post.cs
```diff
@@ -27,7 +27,7 @@ public Post()
         public long? NetVotes { get; set; }
         public decimal Score { get; set; }
         public long Views { get; set; }
-        public long PostTypeId { get; set; }
+        public Enums.PostType PostTypeId { get; set; }
```
asynts, line 30: This should no longer be called `PostTypeId`.

tests/Application.UnitTests/Questions/QuestionsQueryTests.cs
```diff
@@ -0,0 +1,83 @@
...
+                    return question;
+                });
+            for(int i =0;i < 15; i++)
+            {
+                var postA = postsTypeA.ElementAt(i);
+                postA.Id = 0;
+                var postB = postsTypeB.ElementAt(i);
+                postB.Id = 0;
+                _applicationDbContext.Posts.Add(postsTypeA.ElementAt(i));
+                _applicationDbContext.Posts.Add(postsTypeB.ElementAt(i));
```
asynts, line 85: ?

tests/WebApp.IntegrationTests/Pages/QuestionsPageTests.cs
```diff
@@ -0,0 +1,37 @@
...
+            // Arrange
+            var client = _factory.CreateClient();
+
+            // Act
+            var response = await client.GetAsync(url);
+
+            // Assert
+            response.EnsureSuccessStatusCode(); // Status Code 200-299
+            Assert.Equal("text/html; charset=utf-8",
+                response.Content.Headers.ContentType.ToString());
```
asynts: This is a bad check since it would not be incorrect if additional parameters were encoded. `response.ContentType` should be validated instead.

tests/Application.UnitTests/Questions/QuestionsQueryTests.cs
misha130: On insert, the orm automatically changes the id to the new item id. So I have to keep zeroing it so it would actually insert them. Otherwise it ignores them. 
All of this is done so I dont have to write it manually 30 times.

asynts: I missed the <code>postsType<b>A</b>.ElementAt(i)</code> *versus* <code>postsType<b>B</b>.ElementAt(i)</code>, therefor, the confusion. Thanks for clearing this up.

src/Application/Common/Contracts/IRequestHandler.cs
asynts: Actually this might be correct as it is, it depends how the serialization stuff is implemented.

misha130: Its about the async returning only Task.
It was originally like you wrote but I changed it because I want to return Task<TResult> and not TResult

src/Domain/Entities/Post.cs
misha130: I made the PK/FK an enum for convince sake but it still is a foreign key to the PostTypes Table.
Just a matter of convenience

tests/WebApp.IntegrationTests/Pages/QuestionsPageTests.cs
misha130: I cant seem to find `response.ContentType` (mind you response is of type `HttpResponseMessage`)
Also, what do you mean additional parameters?
What additional parameters can you include on the header of ContentType?

asynts: > I can't seem to find `response.ContentType` (mind you the response is of type HttpResponseMessage).

I was a bit sloppy there, the `response.ContentType` worked in controllers I think. `response.Content.Headers.ContentType.MediaType` should work though.

  > What additional parameters can you include on the header of ContentType?

You'd probably never see others in the field, but as far as I know it is allowed to add additional parameters.

src/Domain/Entities/Post.cs
asynts: Hmm. Doesn't it use the builtin "enum" data type? This is the code that made me think that:

https://github.com/codidact/core/blob/d46a6ed857b45a35c252275c7a41c92643562c09/src/Infrastructure/Persistence/ApplicationDbContext.cs#L105

misha130: So just to state the situation:
There is an enum in the db called post_type
There is a table called post_types containing a name & id of the post types
The foreign key of post_type_id and the id of post_types are not of the enum type.

Maybe there is a redundancy in creating this enum inside the db.
Maybe the data types of these columns in the dbs could be the enum? Is this right though?

tests/WebApp.IntegrationTests/Pages/QuestionsPageTests.cs
misha130: 90c3353

misha130:
Well this is what we have so far @luap42 
[Image from Gyazo](https://gyazo.com/f566ba3af1e7df761f9eed4620d006ce)

src/Domain/Entities/Post.cs
asynts: If I understand the situation correctly, the value of id will always be one of the enum values, therefor, setting the type to be that enum seems reasonable.

luap42:
Reviewing this now. I think the seed script is nice, but we'll need some "set up a community" script later, when we come closer to deploying it live.

misha130:
> Reviewing this now. I think the seed script is nice, but we'll need some "set up a community" script later, when we come closer to deploying it live.

Something more complete, yes. Also pull since I added category handling

src/WebApp/Pages/Questions.cshtml
```diff
@@ -0,0 +1,63 @@
...
+                    {
+                        <a href="tag-page.html" class="badge is-tag">@tag.Name</a>
+                        <a href="tag-page.html" class="badge is-tag">@tag.Name</a>
+                    }
+                </div>
+            </div>
+            <div class="item-list--number-value">
+                <div class="item-list--number-label">last active</div>
+                <div class="item-list--number">
+                    @((post.LastModifiedAt ?? post.CreatedAt).ToString("g"))
```
ArtOfCode-: This will localise the date, but it'll localise it to the _server's_ locale not the user's. We should use ISO8601 (`2020-01-01 17:18:19`) and UTC on all server-generated datetimes.

ArtOfCode- (dismissed):

src/WebApp/Pages/Questions.cshtml
misha130: Well I did https://github.com/codidact/core/pull/88/commits/09e2f090ffff9650f2dddd9a8303f8b21a98e5bf which adds a few JS functions to handle this.
The dates are already UTC and I switched "g" to "u".

I am not so sure about the JS because I normally don't do this with pure js and no framework

src/WebApp/Pages/Questions.cshtml
```diff
@@ -0,0 +1,63 @@
+﻿@page "/{category}/questions"
+@model Codidact.Core.WebApp.Pages.Posts.QuestionsModel
+@{
+    ViewData["Title"] = "Questions";
+}
```
luap42, line 8: Showing only posts from the currently chosen category doesn't seem to work yet. Creating a new category and choosing its URL (/meta/Questions) shows the exact same set as the main category (/main/questions)

src/WebApp/Pages/Questions.cshtml
```diff
@@ -0,0 +1,63 @@
...
+<div class="grid has-margin-4">
+    <div class="grid--cell is-flexible">
+        <h1 class="has-margin-0">Q&amp;A posts</h1>
+    </div>
+    <div class="grid--cell">
+        <a href="newpost.html" class="button is-medium has-margin-4 is-filled">New question</a>
+    </div>
+</div>
+<div class="button-list is-gutterless has-margin-top-4 has-float-right">
+    <button class="button is-muted is-outlined is-active">best</button>
```
luap42: These should probably be links changing the ?-part of the URL and hence the query sort order.

src/WebApp/Pages/Questions.cshtml
```diff
@@ -0,0 +1,63 @@
...
+            <div class="item-list--number-value">
+                <div class="item-list--number">
+                    @post.Votes
+                </div>
+                <div class="item-list--number-label">votes</div>
+            </div>
+
+            <div class="item-list--text-value is-primary">
+                <div>
+                    <a href="question.html" class="has-font-size-larger"> @post.Title</a>
```
luap42: This link doesn't work. Should it?

src/WebApp/Pages/Questions.cshtml
```diff
@@ -0,0 +1,63 @@
...
+                <div class="has-padding-top-2">
+                    @foreach (var tag in post.Tags)
+                    {
+                        <a href="tag-page.html" class="badge is-tag">@tag.Name</a>
+                    }
+                </div>
+            </div>
+            <div class="item-list--number-value">
+                <div class="item-list--number-label">last active</div>
+                <div class="item-list--number question-date" data-date="@((post.LastModifiedAt ?? post.CreatedAt).ToString("u"))">
```
luap42: You should also output the date in UTC, so that a valid date is shown for those, who have JS disabled. This can be overwritten by the JS to the local date then.

src/WebApp/wwwroot/js/questions.js
```diff
@@ -0,0 +1,10 @@
+﻿window.addEventListener("load", function () {
```
luap42: Maybe we should put this function to top level and give it a name, so that it can be applied to a class `.live-date`, which can then be updated later live by calling the named function repeatedly (for a more meaningful "x years ago" date).

luap42 (changes requested):

src/WebApp/Pages/Questions.cshtml
misha130: Are you sure you pulled? Right now accessing a category that doesn't exist throws a server exception.
I added an integration test for this too: https://github.com/misha130/core/blob/5a89747ae81698aa70b9bfb01d9035fbf325204a/tests/WebApp.IntegrationTests/Pages/QuestionsPageTests.cs#L37

src/WebApp/Pages/Questions.cshtml
misha130: I'll put the future url there

src/WebApp/Pages/Questions.cshtml
luap42: Yeah. It throws an exception, but it appears, that the question lists aren't adopted (/main/Questions == /meta/Questions). I'll try again. Hold on a second.

src/WebApp/wwwroot/js/questions.js
luap42: (Seeing that you have already done this in a commit: It's probably a better idea to keep the UTC date in a data-argument, because otherwise we loose our original date on an update. So the code should get the date from the [data-date] and then transform it and write it into innerHTML.

src/WebApp/Pages/Questions.cshtml
luap42: Can still repro with latest version:

Screenshot showing two browser windows with one having main and the other meta open and both show the same question list

src/WebApp/wwwroot/js/questions.js
misha130: https://github.com/codidact/core/pull/88/commits/1dbfba00f64ac76ebe4d6539ca2cb5e6c017e944

src/WebApp/Pages/Questions.cshtml
misha130: I changed them to a href with individual sorts

src/WebApp/Pages/Questions.cshtml
misha130: This is odd.
[Image from Gyazo](https://gyazo.com/5950363cb4872f0217ed55b896e656f0)

Can you run the test to see if it fails too?

luap42: Ah. I mean: I created a new category meta in pgAdmin, but didn't assign any post to it.

misha130: Alright, https://github.com/codidact/core/pull/88/commits/2c4ee0543e4f4489aaf33e83604a23956dc83ed8 should resolve this

luap42: Fix confirmed.

src/WebApp/Pages/Questions.cshtml
```diff
@@ -0,0 +1,65 @@
...
+    ViewData["Title"] = "Questions";
+    var query = Model.Query;
+}
+
+<partial name="_ValidationSummary" />
+
+
+<div class="grid has-margin-4">
+    <div class="grid--cell is-flexible">
+        <h1 class="has-margin-0">Q&amp;A posts</h1>
```
luap42: Q&A posts sounds strange. Here should be the category name and/or the post type. So something like "Questions" or "Main Questions".

luap42 (approved):
Looks fine now. Running the tests and then this should be ready for merge.

luap42:
Running the tests fails with this message:

screenshot of error message

> Error Message:
> System.Exception : Category not found:
> Stack Trace
> at Codidact.Core.Application.Questions.Queries.QuestionsQuery.QuestionsQuery.AddFiltersToQuery(QuestionsQueryRequest request, IQueryable`1 questionsQuery) in XXX\CodidactCore\src\Application\Questions\Queries\QuestionsQuery\QuestionsQuery.cs:line 69
   at Codidact.Core.Application.Questions.Queries.QuestionsQuery.QuestionsQuery.Handle(QuestionsQueryRequest request) in XXX\CodidactCore\src\Application\Questions\Queries\QuestionsQuery\QuestionsQuery.cs:line 34
   at Codidact.Core.Application.IntegrationTests.Questions.QuestionsQueryTests.QueryShouldReturnResultsByDefaults() in XXX\CodidactCore\tests\Application.UnitTests\Questions\QuestionsQueryTests.cs:line 30

src/WebApp/Pages/Questions.cshtml
misha130: I fixed this by adding to the query result the category but ideally I should have an area/component/whatever that does this separately from the questions query like some community data.

Also added a short explanation below the category name.

```
    <div class="grid--cell is-flexible">
        <h1 class="has-margin-0">@Model.Result.Category.DisplayName Questions</h1>
        <h3>@Model.Result.Category.ShortExplanation</h3>
    </div>
```

misha130:
Updated these to work with the category

misha130:
> Running the tests fails with this message:

Updated to handle categories

src/Application/Questions/Queries/QuestionsQuery/QuestionsQuery.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        public QuestionsQuery(IApplicationDbContext context, ILogger<QuestionsQuery> logger)
+        {
+            _context = context;
+            _logger = logger;
+        }
+
+        public async Task<QuestionsQueryResult> Handle(QuestionsQueryRequest request)
+        {
+            _logger.LogInformation($"{DateTime.UtcNow.ToString("g")} - Starting to handle request for Questions List");
```
ArtOfCode-: Likewise, use UTC-ISO8601 rather than locale-specific. Server logs will need to be readable in more than one country :)

src/WebApp/Pages/Questions.cshtml
```diff
@@ -0,0 +1,66 @@
...
+    <div class="grid--cell is-flexible">
+        <h1 class="has-margin-0">@Model.Result.Category.DisplayName Questions</h1>
+        <h3>@Model.Result.Category.ShortExplanation</h3>
+    </div>
+    <div class="grid--cell">
+        <a asp-page="questions/new" class="button is-medium has-margin-4 is-filled">New question</a>
+    </div>
+</div>
+<div class="button-list is-gutterless has-margin-top-4 has-float-right">
+    <a asp-page="questions" asp-route-sort="@QuestionsQuerySortType.Best" class="button is-muted is-outlined @(query.Sort == QuestionsQuerySortType.Best?"is-active":"")">best</a>
```
ArtOfCode-: You're missing some spacing in the ternary expression in the `class` attribute here (and likewise in the next two lines).

src/WebApp/Startup.cs
```diff
@@ -124,5 +133,137 @@ private void ApplyDatabaseMigrations(IApplicationBuilder app, ILogger logger)
                 }
             }
         }
+
+        private void SeedDatabase(IApplicationBuilder app, ILogger logger)
```
ArtOfCode-, line 140: DB seed should probably be split off into a separate file/task at some point. Not one for this PR, though.

src/WebApp/wwwroot/js/site.js
```diff
@@ -22,4 +22,32 @@ for (let i = 0; i < headerSlideTriggers.length; i++) {
...
-}
\ No newline at end of file
+}
+
+/*
+ * Date/Time functions
+ ----------------
+ * Utilities to handle dates
+*/
+function utcDateTimeToLocalDisplay(date) {
```
ArtOfCode-, line 32: Have we got Babel preprocessing set up yet? If not, we should make that a priority. If we do, this file should be reviewed and updated to ES6 - notably, `const` and `() => { }`.

ArtOfCode- (changes requested):
This is mostly a syntax/style review, rather than techniques. Couple of minor changes, but mostly good to go.

src/WebApp/wwwroot/js/site.js
misha130: Not yet basically. Keeping it ES3 so far

src/WebApp/Pages/Questions.cshtml
misha130: Could've been something like:
```
    @foreach (var sortType in (QuestionsQuerySortType[]) Enum.GetValues(typeof(QuestionsQuerySortType)))
    {
        <a asp-page="questions" asp-route-sort="@sortType" class="button is-muted is-outlined @(query.Sort == sortType ? "is-active" : "")">@sortType</a>
    }
```

But for now just formatted the condition

src/WebApp/wwwroot/js/site.js
ArtOfCode-: Works for now. Just bear in mind that the longer it takes to get it set up, the more JS there'll be to rewrite when it's ready :)

ArtOfCode- (approved):

luap42:
@misha130 please merge develop into this PR and resolve the conflicts, theb we can merge this PR.

misha130:
> @misha130 please merge develop into this PR and resolve the conflicts, theb we can merge this PR.

Ok done

Issue #63, opened 2020-02-20
Implement localization 


misha130:
Per the forum thread https://forum.codidact.org/t/please-bake-in-localization-from-the-beginning/605 we should "bake in localization".

Dotnet supports this as part of its framework:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/localization?view=aspnetcore-3.1

Should at least create a Resources file and implement a standard of using it.

Note: Auth project should have this issue too

misha130:
Closed by #89

PR #89, opened 2020-04-14
Add localization

misha130:
### Pull Request summary

- Adds Translations
- Adds a basic culture ("en") with a default

https://github.com/codidact/core/issues/63

### Pull Request Tasklist

### DIFF
--- a/src/WebApp/Startup.cs
+++ b/src/WebApp/Startup.cs
@@ -83,8 +83,7 @@ namespace Codidact.Core.WebApp
             services.Configure<RequestLocalizationOptions>(
              opts =>
             {
-                var settingSupportedCultures = Configuration.GetSection("Localization:Cultures")
-                                                .GetChildren().Select(child => child.Value);
+                var settingSupportedCultures = Configuration.GetSection("Localization:Cultures").Get<string[]>();
                 if (settingSupportedCultures == null || !settingSupportedCultures.Any())
                 {
                     throw new CultureNotFoundException("No supported cultures found in settings");
--- a/src/WebApp/Startup.cs
+++ b/src/WebApp/Startup.cs
@@ -123,8 +123,7 @@ namespace Codidact.Core.WebApp
 
             app.UseRouting();
 
-            var localizationOptions = app.ApplicationServices.GetService<IOptions<RequestLocalizationOptions>>().Value;
-            app.UseRequestLocalization(localizationOptions);
+            app.UseRequestLocalization();
 
             app.UseAuthentication();
             app.UseAuthorization();
--- a/src/WebApp/Startup.cs
+++ b/src/WebApp/Startup.cs
@@ -14,6 +14,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
 using System;
 using System.Collections.Generic;
 using System.Globalization;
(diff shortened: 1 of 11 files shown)


asynts (commented):
This looks good, but I disagree with the logic that chooses which culture is used to represent the page.

*(The following is my opinion.)*

My understanding was that the canonical communities are in English. Therefor, there should not possible to change the interface language to something else because this would encourage people to create posts in that language.

However, there would be communities where this would make sense. Maybe a Spanish community or something of the like. 

The logic that chooses the culture is determined by the [`RequestLocalizationOptions.RequestCultureProviders`][1] property. In my opinion the defaults are not what we want, there should be two implementations of `IRequestCultureProvider`:

  - `CommunityConstantCultureProvider` this would set the culture to a constant which we would configure with the `appsettings.json` file. This provider would be used for the English communities. 

  - `CommunityChoiceCultureProvider` this would allow a user to choose a language from a list of allowed languages. For a Spanish community we would allow the choices "English" and "Spanish" for example.

## Addendum

Actually, both could be done in a single implementation. Just read an array from the configuration file and if there is only one option, make it constant.

  [1]: https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.builder.requestlocalizationoptions.requestcultureproviders?view=aspnetcore-3.1#Microsoft_AspNetCore_Builder_RequestLocalizationOptions_RequestCultureProviders

misha130:
> Actually, both could be done in a single implementation. Just read an array from the configuration file and if there is only one option, make it constant.

So in general if I switch the configuration to read app.settings for the supported cultures it would be okay? Without implementing a `IRequestCultureProvider`? Or do I need the impl to stop people from switching on the constant

asynts:
> So in general if I switch the configuration to read app.settings for the supported cultures it would be okay? Without implementing a IRequestCultureProvider? Or do I need the impl to stop people from switching on the constant

That's even simpler, that should work.

src/WebApp/Startup.cs
```diff
@@ -97,6 +123,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env, ILogger<
 
             app.UseRouting();
 
+            var localizationOptions = app.ApplicationServices.GetService<IOptions<RequestLocalizationOptions>>().Value;
+            app.UseRequestLocalization(localizationOptions);
```
asynts: I think you can just use the following, it should pick up the options automatically. (I can't test this right now.)

~~~csharp
app.UseRequestLocalization();
~~~

src/WebApp/Startup.cs
```diff
@@ -73,7 +77,29 @@ public void ConfigureServices(IServiceCollection services)
...
                 .AddRazorRuntimeCompilation()
-                .AddFluentValidation();
+                .AddFluentValidation()
+                .AddViewLocalization(options => options.ResourcesPath = "Resources");
+
+            services.Configure<RequestLocalizationOptions>(
+             opts =>
+            {
+                var settingSupportedCultures = Configuration.GetSection("Localization:Cultures")
+                                                .GetChildren().Select(child => child.Value);
```
asynts: I can't test this right now, but it should be possible to do this as follows:

~~~csharp
var settingSupportedCultures = Configuration.GetValue<string[]>("Localization:Cultures");
~~~

misha130: Nope. It was the first thing I tried too.
Making a class is an overkill for now

src/WebApp/Startup.cs
asynts: I've tested this, it is able to pick up the options from the services collection.
```suggestion
            app.UseRequestLocalization();
```

src/WebApp/Startup.cs
asynts: I've tried this now and you are correct, but the following works:

```suggestion
                var settingSupportedCultures = Configuration.GetSection("Localization:Cultures").Get<string[]>();
```

asynts (approved):
I've tested this on my machine and it works. (There was an issue but it was unrelated to this pull request, I created #90 for this.)

src/WebApp/Startup.cs
misha130: Very cool. Couldn't figure it out myself

luap42 (approved):

PR #92, opened 2020-04-15
Typescript support

misha130:
### Pull Request summary

Transpiles JavaScript code to support older browsers.
Current target is set to ES5.

To run from CLI need to:

- Install node
- Execute npm i
- run script "npm run serve" to serve
- run script "npm build" to build

For Visual Studio this is automatic


src/WebApp/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "codidact.core",
+  "version": "1.0.0",
+  "description": "Codidact Platform",
+  "scripts": {
+    "tsc": "tsc",
```
asynts: I am not super familiar with npm, but shouldn't this be `npx tsc`?

src/WebApp/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
asynts, line 1: We do not plan to make this an npm package, therefor, only the `devDependencies` and `scripts` fields are required. In my opinion the rest should be removed because they might become outdated.

src/WebApp/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "noImplicitAny": false,
+    "noEmitOnError": true,
+    "removeComments": false,
+    "sourceMap": true,
+    "target": "es5"
```
asynts, line 7: I thought we were targeting ES6? I do not know where I picked that up though.

misha130: If that was the case then there would be no point in adding TypeScript.
Just adding this instead of babel basically.

src/WebApp/package.json
misha130: True, I have ts installed globaly so this is why it worked for me

src/WebApp/tsconfig.json
asynts: I mean TypeScript does a lot more than just babel. Not sure if I am pinging the right person here, but @mattjbrent you do a lot of frontend stuff right? ES5, ES6 or TypeScript?

bas-ie: Just browsing past... TS will give you some type safety in the front-end (at build-time, not run-time) and good Intellisense from VS Code. It's pretty useful, with great editor support. In terms of the build target though, it probably comes down to which browsers you're planning to support. `es5` might be safer.

luap42: Let's use es5 for the built JavaScript.

Daksh14 (approved):

PR #93, opened 2020-05-06
Documented code

ghost:
### Pull Request summary

Documented the code for readability

None.

### Pull Request Tasklist

- [ ] Pull request references an issue
- [ ] Branch follows contributing guidelines format
- [X ] Pull request targets the `develop` branch

### DIFF
--- a/src/WebApp/Program.cs
+++ b/src/WebApp/Program.cs
@@ -7,12 +7,15 @@ namespace Codidact.Core.WebApp
 {
     public class Program
     {
+        // Task Main makes an instance of
+        // HostBUilder and runs it asynchronously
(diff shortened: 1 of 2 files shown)


misha130:
Though this is actually super helpful some points:

- Use xmldoc to comment methods/classes/props (write /// and the IDE completes the rest I think)
- Some of the comments in startup.cs are not accurate as in it doesn't really configure services or cookies
- If you are commenting code inside of a method it means this code should be moved to a separate method (as in it smells) the calls should be self explanatory with their names.

There are many many many files that I missed the documentation on props/methods/classes though and it would be great if it was addressed with xmldoc

